PR #3, opened 2017-12-15
Adwords: MCC Account Requirement

erinkcochran87:
This PR adds info to the AdWords setup requirements about linking MCC accounts prior to connecting Stitch.

This step is necessary because AdWords API access is not available on advertiser accounts. Access is only granted to managed accounts, making the connection necessary to extract data via Stitch.


erinkcochran87:
@briansloane - could I get your 👀  on this? [Here's the preview version](https://deploy-preview-3--stitchdocs.netlify.com/docs/integrations/saas/google-adwords#setup-requirements).

PR #4, opened 2017-12-15
SendGrid: PKs & `dropped` events

erinkcochran87:
This PR adds a subsection to SendGrid’s **Replication** section that details why some events may not be persisted.

This is a combo of how SendGrid’s API occasionally functions and the way Stitch is built to handle Primary Keys.


erinkcochran87:
@bi1yeu approved copy via internal ticket - merging.

PR #5, opened 2017-12-21
Remove referral program doc

khalkyer:
Beginning 12/21 we are discontinuing the referral program. This doc can be removed.


erinkcochran87:
I'm not sure why the Netlify preview isn't working despite a manual re-triggering, but I checked this out locally and everything looks fine. Thanks for handling this, @khalkyer :)

PR #31, opened 2018-03-06
Mixpanel: Export and Engage have nested data

matt-zach-s:
The response from [Mixpanel API](https://mixpanel.com/help/reference/exporting-raw-data#export-api-reference) includes an object that includes all of the attributes of the given event. This results in flattening of records.

See:

"One event per line, sorted by increasing timestamp. *Each line is a valid JSON object* although the return itself is valid JSON but instead JSONL."


erinkcochran87:
@matt-zach-s ping ^^^

matt-zach-s:
Hey @erinkcochran87 

This is intended. The [Engage endpoint that Stitch uses](https://mixpanel.com/help/reference/data-export-api#people-analytics) also has nested structures. I can update the title of the PR to mention this.

PR #40, opened 2018-03-28
Update Schema of mixpanel_export

dylan-stitch:
We listed the field as _sdc_record_hash, though this field is loaded as _rjm_record_hash.


erinkcochran87:
@dylan-stitch This is true of newly created Mixpanel integrations?

dylan-stitch:
@erinkcochran87 yup. See: https://github.com/stitchdata/sourcerer/blob/master/sourcerer/src/clojure/sourcerer/apis/mixpanel.clj#L70

PR #41, opened 2018-04-03
Singer plug

matt-zach-s:
I have modified the wording here in an attempt to encourage the reader to explore the Singer specs as a means of interacting with the Stitch Import API. That is instead of writing custom import api scripts.


erinkcochran87:
Closing - I've included this in the re-write of the IAPI docs.

PR #44, opened 2018-04-11
Facebook Ads/v1.4.0

erinkcochran87:
This PR:

- Adds the `ads_insights_region` table to the FB Ads schema (https://github.com/singer-io/tap-facebook/pull/40)
- Adds instructions and info about configurable attribution windows
- Adds the `attribution_window` attribute to Facebook's Connect API Source Form Property


erinkcochran87:
@briansloane I added the changes we talked about - [care to take a look before I merge this](https://deploy-preview-44--stitchdocs.netlify.com/docs/integrations/saas/facebook-ads#select-attribution-window)?

briansloane:
This looks great. I like the callout box for explaining the situation with 2 different values.

One noteworthy change, I think on historical syncs we still use the attribution window on top of the start_date, so we'll actually go back N days from your start_date on the historical sync as well.

erinkcochran87:
Oh interesting - that does make sense, but I hadn't thought it would happen there.

I just reset the keys in one of my test integrations; I'll check out the logs in the morning to confirm. This will be good to ship out tomorrow either way :)

PR #73, opened 2018-05-25
Zendesk v1.0

erinkcochran87:
This PR adds documentation for the new version of the Zendesk integration.

### DIFF
--- a/_data/taps/versions/zendesk.yml
+++ b/_data/taps/versions/zendesk.yml
@@ -1,0 +1,16 @@
+# -------------------------- #
+#      ZENDESK VERSIONS      #
+# -------------------------- #
+
+latest-version: "1.0"
+
+released-versions:
+  - number: "1.0"
+    date-released: "TBD"
+    deprecated: false
+    # deprecation-date: "n/a"
+
+  - number: "15-10-2015"
+    date-released: "October 15, 2015"
+    deprecated: true
+    # deprecation-date: "n/a"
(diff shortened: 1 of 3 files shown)


HarrisonMarcRose:
Hmmm. I just looked at the docs. Actually, I'm not sure. Maybe I changed
that and shouldn't have. We should talk to Dan. It would make sense that we
are using that but I don't know for sure.

Harrison | Lead QA Engineer

erinkcochran87:
@dmosorast - Can you confirm for us which API the `organizations`, `ticket_metrics`, `tickets`, and `users` tables use to extract data? The project plan for this says it's the Incremental Export API, but we're unsure.

Based on what I see in the tap code and what's in Zenpy, it looks like we're using the Incremental Export API for these objects:

- `organizations` = [Tap code](https://github.com/facetoe/zenpy/blob/50cb95fcc17751d539c5953216ef57f964b77913/zenpy/lib/endpoint.py#L424) // [Zenpy code](https://github.com/facetoe/zenpy/blob/master/zenpy/lib/endpoint.py#L424)
- `tickets` = [Tap code](https://github.com/singer-io/tap-zendesk/blob/e6ce54c6449873bacf3a555446ce99979eecddc1/tap_zendesk/streams.py#L141) // [Zenpy code](https://github.com/facetoe/zenpy/blob/master/zenpy/lib/endpoint.py#L455)
- `users` = [Tap code](https://github.com/singer-io/tap-zendesk/blob/e6ce54c6449873bacf3a555446ce99979eecddc1/tap_zendesk/streams.py#L129) // [Zenpy code](https://github.com/facetoe/zenpy/blob/master/zenpy/lib/endpoint.pyL474)

`ticket_metrics` I'm unsure of from looking at the code, as the tap seems to reference just the primary endpoint:

```
# tap/streams.py:

ticket_metrics = self.client.ticket_metrics()

# zenpy/endpoint.py:

 ticket_metrics = PrimaryEndpoint('ticket_metrics')
```

PR #76, opened 2018-05-30
GitHub: v1

erinkcochran87:
This PR adds documentation for the [community-supported GitHub integration](https://github.com/singer-io/tap-github).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Gemfile.lock
 .Trashes
 ehthumbs.db
 Thumbs.db
+*~
--- a/_includes/integrations/templates/schemas/table-schemas.html
+++ b/_includes/integrations/templates/schemas/table-schemas.html
@@ -50,21 +50,30 @@
     2. Captures attributes marked as Replication Keys -->
     {% if table.replication-method contains "Incremental" %}
 
-        {% capture clean-replication-keys %}
-            {% for attribute in table.attributes %}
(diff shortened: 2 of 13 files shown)


erinkcochran87:
@KAllan357 This is looking great - just a few questions:

- `pull_requests` and `reviews` are listed as Full Table, but it looks like there might be a state used in the tap code (ex: [here](https://github.com/singer-io/tap-github/blob/master/tap_github.py#L138)). Are they extracted using Full Table from the API, and then persisted as Incremental? Or no?
- `commits` is listed as Incremental, but I didn't see an obvious Replication Key for the table. Is new/updated data identified using a query filter?
- Does the user need to define any additional scopes when creating the access token, or are their permissions to the project enough?
- Is this a one repository : one integration setup? Or can users enter multiple repos into the Integration Settings page?

KAllan357:
1. They write state, but don't use it. So they are extracted and persisted as Full Table.
2. Right the replication key we use is `since` which is the name of the query param. In this case, there's no field on the table that relates to this - its just commits that have been added since the last time we checked.
3. When you create the access token you can choose what permissions it has. I think the read repo one is sufficient. I don't think any others apply though I could be wrong. Unfortunately, I don't have a good way to test.
4. At the moment, its one repository to one integration.

erinkcochran87:
Cool beans - thanks! I'm gonna test out the token permissions and then, depending on what I find, I think this will be good to go.

PR #84, opened 2018-06-07
Listrak/v1

ajmers:
This PR adds documentation for the Listrak integration.


erinkcochran87:
@KAllan357 The rendered preview [is available here](https://deploy-preview-84--stitchdocs.netlify.com/docs/integrations/saas/listrak).

_saas-integrations/listrak/listrak-latest.md
```diff
@@ -0,0 +1,90 @@
...
+certified: false
+
+historical: "1 year"
+frequency: "30 minutes"
+tier: "Free"
+status-url: 
+icon: /images/integrations/icons/listrak.svg
+whitelist:
+  tables: true
+  columns: false
```
KAllan357: I'm not aware of how we can confirm (as we don't have a test account), but I think we do support propert-selection here so this should be true I think.

KAllan357:
👍

_saas-integrations/listrak/listrak-latest.md
erinkcochran87: Isn't there something that's added to the tap itself to enable field selection?

KAllan357: Yes. Upon closer inspection, it looks like everything is just set to "Automatic" so I think this is okay.

PR #87, opened 2018-06-11
Google CloudSQL MySQL: Binlog

erinkcochran87:
This PR adds instructions for setting up binlog for Google CloudSQL MySQL databases.


erinkcochran87:
@ccapurso Could you give this a once-over to make sure all the required steps for setting up binlog are in here? [This is the only section in the doc](https://deploy-preview-87--stitchdocs.netlify.com/docs/integrations/databases/google-cloudsql-mysql#server-settings) that really needs reviewed - everything else is pretty much the same as it was before.

The one thing I'm unclear on are what privileges the user performing [step 2.2](https://deploy-preview-87--stitchdocs.netlify.com/docs/integrations/databases/google-cloudsql-mysql#define-database-parameters) needs to have. I saw in Google's docs that they don't have a `SUPER` privilege, which is required for [vanilla MySQL configuration](https://dev.mysql.com/doc/refman/5.6/en/privileges-provided.html#priv_super), but their docs didn't say what, if anything, was required to do this.

PR #106, opened 2018-07-17
Append-Only example: Rep Key inclusivity with updated_at

matt-zach-s:
The original does not take into account the fact that during append-only replication, records which have a replication key value greater than or equal to the previous bookmark will also be replicated.


erinkcochran87:
@matt-zach-s Thanks for bringing this to my attention. I did wind up re-writing this example (mostly because the original wasn't up to my current standards) - you can take a look at my commit (https://github.com/stitchdata/docs/pull/106/commits/62f777e1072380b780c3a9a7244b3838ea37be8f) to compare what I did to what you submitted originally.

The `updated_at` values were a little funky from table to table - just something to keep in mind for next time!

PR #113, opened 2018-07-27
Deleted records

erinkcochran87:
This PR improves the documentation for deleted record handling, and adds new docs for `_sdc_deleted_at`.

### DIFF
--- a/_data/dataloading/deleted-records.yml
+++ b/_data/dataloading/deleted-records.yml
@@ -81,7 +81,12 @@ log-based-incremental:
 
       **Note**: Identifying hard deleted records is dependent on how the record is deleted, as not all event types are recorded by databases in binary logs. For example: If a record is deleted using `TRUNCATE` in a MySQL database, Stitch will not detect the deletion.
 
-      Refer to the [MySQL]() and [PostgreSQL]() documentation for more details.
+      The following event types will be detected by {{ replication-method-display-name }} for MySQL- and PostgreSQL-backed database integrations:
+
+      - `INSERT`
+      - `UPDATE`
+      - `DELETE`
+
     columns:
       - name: "id"
       - name: "name"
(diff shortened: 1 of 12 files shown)


_data/dataloading/deleted-records.yml
```diff
@@ -0,0 +1,163 @@
...
+
+
+#------------------------------------#
+#        LOG-BASED INCREMENTAL       #
+#------------------------------------#
+
+binlog-replication:
+
+#------------------------------------#
+#       LOG-BASED, HARD DELETES      #
```
dylan-stitch: My understanding was that only the PK value and the _sdc fields would remain when we capture a hard-deletion event with log based incremental. If that's true, it might make sense to add an additional field or two for these examples to show that we clear out those values when we mark the record as deleted, where only the PK and the _sdc fields would remain post delete.

_data/dataloading/deleted-records.yml
```diff
@@ -0,0 +1,163 @@
...
+                updated-at: "2018-07-28 04:15:00"
+
+              - id: "2"
+                updated-at: "2018-07-27 12:05:00"
+
+              - id: "3"
+                updated-at: "2018-07-28 04:07:00"
+
+#------------------------------------#
+#       FULL TABLE REPLICATION       #
```
dylan-stitch: The one thing I noticed was that this doesn't have information specific to how full table works for BigQuery (eg that it depends on the data source). Would that warrant it's own doc page?

_data/dataloading/deleted-records.yml
erinkcochran87: Oh really? I didn't know that was the case here. I'll see if we have a test database I can play with to verify this.

_data/dataloading/deleted-records.yml
erinkcochran87: Oh right - in that it just appends the table in full after each job, right?

erinkcochran87: Though, I believe what would happen in BQ would happen in any table that is updated using Append-Only... so this might mean I add another tab specific to Append-Only (with some notes) and do additional examples for each delete type.

dylan-stitch: It appends in full after each job with the exception of database integrations and salesforce.

erinkcochran87:
I think I'm going to break this out into its own doc, instead of having it as a section in the Replication Methods doc. That'll give me some room to talk about the weirdness of Append-Only destinations.

erinkcochran87:
Finally done - merging.

PR #114, opened 2018-07-30
Zendesk v1: Add ticket_forms stream

nick-mccoy:
I recently added the `ticket_forms` stream to tap-zendesk.  This is the corresponding documentation PR for it.  

A few questions:  

- Do we include foreign key distinctions for streams that the tap doesn't currently support?  (e.g. the `restricted_brand_ids` attribute)
- When I added the `ticket_forms` stream, I bumped the version to 1.1.0.  Should this be reflected in the `version` in this stream's doc or does that refer to a different version?

Singer PR: https://github.com/singer-io/tap-zendesk/pull/15

### DIFF
--- a/_integration-schemas/zendesk/ticket_forms.md
+++ b/_integration-schemas/zendesk/ticket_forms.md
@@ -88,8 +88,6 @@ attributes:
       - name: "value"
         type: "integer"
         description: "The ID of the brand."
-        foreign-key: true
-        table: "brands"
   
 
     
--- a/_integration-schemas/zendesk/ticket_forms.md
+++ b/_integration-schemas/zendesk/ticket_forms.md
@@ -88,9 +88,4 @@ attributes:
       - name: "value"
         type: "integer"
         description: "The ID of the brand."
-  
-
-    
-
-
-
+---
--- a/_integration-schemas/zendesk/ticket_forms.md
+++ b/_integration-schemas/zendesk/ticket_forms.md
@@ -1,0 +1,98 @@
+---
+tap: "zendesk"
+version: "1.0"
+
+name: "ticket_forms"
+doc-link: https://developer.zendesk.com/rest_api/docs/core/ticket_forms
+singer-schema: https://github.com/singer-io/tap-zendesk/blob/master/tap_zendesk/schemas/ticket_forms.json
+description: |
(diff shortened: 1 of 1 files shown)


erinkcochran87:
@nick-mccoy 

Re: 1: Not currently. Do you think we should? It feels strange to me to point to a foreign key for a table the user doesn't have access to.

Re: 2: Nah - right now I'm only reflecting major version bumps in the documentation. So `1.0` just refers to version `1.x.x`. I figure for anything that isn't a breaking change or doesn't require a migration, this is okay for now since users on the version will have access to it anyway.

nick-mccoy:
Thanks, @erinkcochran87!  Agreed on (1), totally makes sense and I removed the foreign key distinction for that attribute.

_integration-schemas/zendesk/ticket_forms.md
```diff
@@ -0,0 +1,98 @@
...
+  - name: "restricted_brand_ids"
+    type: "array"
+    description: "IDs of all brands that this ticket form is restricted to."
+    array-attributes:
+      - name: "value"
+        type: "integer"
+        description: "The ID of the brand."
+        foreign-key: true
+        table: "brands"
+
```
erinkcochran87: This table won't show up in Zendesk's docs because it doesn't have a closing `---` - [here's the preview, where `ticket_forms` is missing](https://deploy-preview-114--stitchdocs.netlify.com/docs/integrations/saas/zendesk#schema).

In Jekyll, files have to have starting and closing `---` (see line one in this file). These mark the beginning and end of a file's Frontmatter section (which is what this file is in its entirety) and must be present for Jekyll to read the file.

If you add `---` on line 93, this'll be good to go!

(Also, sorry to kick this back to you for such a minor thing - I'm trying to break the habit of fixing people's PRs, and you're the first 😉)

nick-mccoy: Ah, ok, makes sense!  Just committed the change

And no worries on kicking the PR back, we're trying to get in to the same habit for Singer community based PRs, so I'm glad to see it in practice :)

erinkcochran87:
Awesome - thanks for adding that line in!

I'm also going to tack Zendesk's changelog to this PR just to ensure this update doesn't slip through the cracks. I plan on this feature going out in the next week or so.

PR #124, opened 2018-08-10
Integration schema redo

erinkcochran87:
This PR focuses on cleaning up the code for integration schema templates. It also adds in display logic for foreign keys that are contained within nested structures, such as an `object` or an `array`.


erinkcochran87:
For **docs about docs**: The `foreign-key` parameter for table attributes is now `foreign-key-id`. This is use to uniquely identify a mapping in `foreign-keys.yml`, allowing lists of columns to be single-sourced and re-used across an integration's entire schema.

PR #126, opened 2018-08-14
Add more options for using yaml generator

dmosorast:
While looking at updating the docs for the tap-harvest V2 upgrade, I realized that the schema generator will take a single file, or a catalog, but this doesn't work for taps with a large number of schemas and no discovery mode (so no catalog).

This PR adds the ability to specify multiple files, or a directory, and it will recursively search the directory for all files that match `*.json` and generate a yaml file for each based on the file name.

Here's the new usage text:

```
Usage: schema_to_yaml.rb schema_file [schema_file2 ...]
       schema_to_yaml.rb catalog_file
       schema_to_yaml.rb /path/to/schemas
 * If a schema has $ref values, the ref will be resolved relative to the location of the file being processed.

Options:
       -o file_path    Specifies an output path for the generated files (relative to the current directory, or absolute)
```

### DIFF
--- a/scripts/schema_to_yaml.rb
+++ b/scripts/schema_to_yaml.rb
@@ -4,16 +4,19 @@ require 'json'
 require 'yaml'
 
 if ARGV.length < 1
-  puts "Usage: schema_to_yaml.rb schema_file [schema_file2 ...]"
-  puts "       schema_to_yaml.rb catalog_file"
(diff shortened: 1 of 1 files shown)


erinkcochran87:
I know there are other changes to this script in the works, but in the interest of still being able to work efficiently while that's wrapped up, I'm gonna merge this. After being able to zip through all the schemas at once, I NEVER WANT TO GO BACK.

PR #135, opened 2018-08-21
Zendesk/v1 - Add repo-url & update Singer flag

dylan-stitch:
changes `singer: false` to `singer: true`


erinkcochran87:
I also added the repo URL in here, since we chatted about it earlier today.

PR #142, opened 2018-08-27
Postgres/v1: Remove v10 notice from binlog setup instructions

dylan-stitch:
removes indication that postgresql 10 is not supported, as this support has been added.


erinkcochran87:
Good catch - I removed this in the Prereqs section the other day, but apparently missed this. Thank you!

PR #155, opened 2018-09-13
Closeio/v1

erinkcochran87:
This PR adds documentation for the new version (v1) of the Close.io integration.


_data/taps/versions/closeio.yml
```diff
@@ -0,0 +1,18 @@
+# -------------------------- #
+#   [INTEGRATION] VERSIONS   #
+# -------------------------- #
+
+latest-version: "1.0"
```
timvisher: Do we say `1.0` when we mean `1.x`? The latest version is `1.5.0`

erinkcochran87: Yes. I'm not including minor versions/patches in the docs - I'd have to constantly be updating these, and from a customer perspective, only major version changes will really affect them.

PR #165, opened 2018-09-25
Feature/schema2md clojure script

dmosorast:
This PR updates the schema to YAML script.


scripts/tap-generate-docs/src/tap_generate_docs.clj
```diff
@@ -51,14 +51,13 @@
...
-                               (if ((set other-properties) "array")
-                                 (throw (ex-info "Currently cannot handle a type with object _and_ array present"
-                                                 {:property property}))
-                                 other-properties))
-            converted-other-properties [(convert-multiary-type schema ["value"
-                                                                       (assoc property-json-schema-partial
-                                                                              "type"
-                                                                              other-properties)])]
+                               other-properties)
+            converted-other-properties (if (= '("null") other-properties)
```
timvisher: Why not `(if (= ["null"] other-properties))` here?

scripts/tap-generate-docs/src/tap_generate_docs.clj
```diff
@@ -48,16 +50,15 @@
       (convert-object-properties schema (property-json-schema-partial "properties"))
       ;; Alternate
       (let [object-properties (convert-object-properties schema (property-json-schema-partial "properties"))
-            other-properties (let [other-properties (if (string? item-type)
-                                                      [item-type]
-                                                      (filter (partial not= "object") item-type))]
-                               other-properties)
+            other-properties (if (string? item-type)
+                               '(item-type)
```
timvisher: Why not `[item-type]`?

scripts/tap-generate-docs/src/tap_generate_docs.clj
```diff
@@ -36,7 +36,9 @@
   [schema properties]
   {:pre [(not (contains? properties "type"))]}
   (sort-by #(get % "name")
-           (map (partial convert-multiary-type schema) properties)))
+           (reduce (fn [acc p] (concat acc (convert-multiary-type schema p)))
+                   []
+                   properties)))
```
timvisher: This is very odd to me. `map` produces a seq so reducing into a `vector` seems weird. Was the problem that it wasn't being realized? `doall` should work there.

scripts/tap-generate-docs/src/tap_generate_docs.clj
dmosorast: I believe we observed that the value at this point came through as a list, but looking at it now, it seems like it could be a vector. I'm still kind of fuzzy at the distinction between the two with regard to equality.

scripts/tap-generate-docs/src/tap_generate_docs.clj
dmosorast: Since filter returns a list, it seemed appropriate to also return a list in the literal case.

scripts/tap-generate-docs/src/tap_generate_docs.clj
dmosorast: Nm, we changed it to a list in a later commit. See other comment.

scripts/tap-generate-docs/src/tap_generate_docs.clj
dmosorast: The reason for the reduce is that we pulled the literal vector of `[(convert-multiary-type ...)]` into the return value of `convert-multiary-type`, so it's reducing into a vector because that's the return value's structure.

erinkcochran87:
I'm really excited to test this out! Can one of y'all put some instructions for how to use this in here, please? @dmosorast @timvisher

timvisher:
@erinkcochran87 We'll iron out the last bits of the changes @dmosorast and @nick-mccoy made and then I can definitely walk you through setting it up. :)

scripts/tap-generate-docs/src/tap_generate_docs.clj
```diff
@@ -171,7 +178,9 @@
         (do
           (println (str "Null unary type passed for property"
                         {:property property}))
-          [])
+          {"name" property-name
+           "type" ""
```
timvisher: Wouldn't this be `"type" "null"`?

scripts/tap-generate-docs/test/tap_generate_docs_test.clj
```diff
@@ -279,13 +288,11 @@
...
-#_(testing "Null types"
-    (is (thrown? clojure.lang.ExceptionInfo
-                 (convert-multiary-type nil ["a_null" {"type" "null"}])))
-    (is (thrown? clojure.lang.ExceptionInfo
-                 (convert-multiary-type nil ["a_null" {"type" ["null"]}]))))
+(testing "Null types"
+    (is (not (thrown? clojure.lang.ExceptionInfo
+                  (convert-multiary-type nil ["a_null" {"type" "null"}]))))
+    (is (not (thrown? clojure.lang.ExceptionInfo
+                  (convert-multiary-type nil ["a_null" {"type" ["null"]}])))))
```
timvisher: I think it would be better for us to test the actual return value these should be now (`nil?`, a `null` type doc partial?), rather than that they simply don't throw.

scripts/tap-generate-docs/src/tap_generate_docs.clj
```diff
@@ -21,6 +21,13 @@
   (or (= {} property-json-schema-partial)
       (string? (property-json-schema-partial "type"))))
 
+(defn type-declaration?
+  "A valid type declaration is either a string or a list. Never a map,
+   which covers the rest of JSON schema."
+  [sub-schema]
+  (or (seq? sub-schema)
```
timvisher: I'm unsure if this will do exactly what you think it will do. 

```
user=> (seq? [:foo :bar])
false
user=> (seq? {:foo :bar})
false
user=> (seq? (seq [:foo :bar]))
true
user=> (seq? (seq {:foo :bar}))
true
```

I think instead you want `sequential?`

```
user=> (sequential? {:foo :bar})
false
user=> (sequential? [:foo :bar])
true
```

but be cautious that would be true for anything that was mapped over.

```
user=> (sequential? (map identity {:foo :bar}))
true
```

erinkcochran87:
@timvisher @dmosorast What's happenin' with this? Can it be merged/closed?

dmosorast:
@erinkcochran87 Ah! Sorry, I missed the comments on the latest change. I'll put it on my list to clean up. 😅

dmosorast:
@timvisher I made the changes requested as part of the latest Stripe docs generation. Are these good?

@erinkcochran87 Are you okay if we merge this?

We have [a card](https://trello.com/c/vZFYEhTh/351-internal-merge-the-docs-generator) to get this merged, and not sure why it's waiting right now. Given the go ahead. I'll just merge it. Shouldn't affect the rest of the repo at all, iiuc.

erinkcochran87:
@dmosorast I feel like I've waited so long for this day! 🎉 🎉 🎉  Merge away, friend!

dmosorast:
Thanks! Github is blocking merge for me with `You’re not authorized to merge this pull request.`.

It's either as a result of the "Requested changes" status, or something with my permissions on the repo?

timvisher:
Good from my end!

dmosorast:
Great! @erinkcochran87 I guess I don't have permissions to merge, but you should be good to merge this whenever you like! :shipit:

PR #183, opened 2018-10-22
Removes current-step-hints documentation from connect-api docs

iterati:
current-step-hints is no mo


erinkcochran87:
It looks like `current_step_hints` weren't removed from the other code examples before this was closed. I'll take care of it now.

PR #205, opened 2018-11-21
Azure SQL DW: QA review

HarrisonMarcRose:
This PR makes some minor corrections to the docs for the Azure SQL DW destination.

### DIFF
--- a/_data/destinations/reference/microsoft-azure.yml
+++ b/_data/destinations/reference/microsoft-azure.yml
@@ -74,7 +74,7 @@ replication-info:
 
 object-name-limit-info:
   case-sensitivity: &case-sensitivity "Insensitive"
-  tables: &table-name-length "128 characters"
+  tables: &table-name-length "112 characters"
   columns: &column-name-length "128 characters"
 
 # -------------------------- #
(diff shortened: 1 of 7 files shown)


_data/destinations/reference/microsoft-azure.yml
```diff
@@ -185,7 +185,7 @@ attributes:
     description: |
       {{ destination.display_name }} destinations do not have native support for Primary Keys.
 
-      To ensure existing rows can be updated in incrementally-replicated tables, Stitch will create an `sdc_primary_keys` table in every integration schema. This table contains the Primary Key data for every table in the given schema, which Stitch will use to de-dupe rows during loading.
+      To ensure existing rows can be updated in incrementally-replicated tables, Stitch will create an `_sdc_primary_keys` table in every integration schema. This table contains the Primary Key data for every table in the given schema, which Stitch will use to de-dupe rows during loading.
```
erinkcochran87: And for every other `_sdc_*` hereafter:

I thought we determined that object names couldn't begin with underscores?

HarrisonMarcRose: I saw you had that in the table names section too and wasn't sure what we do if the customer had a table starting with underscore. At least for our reject and ok table they definitely start with an underscore.

PR #209, opened 2018-12-03
Replication methods overhaul

erinkcochran87:
This PR overhauls the Replication Methods documentation.


_replication/replication-methods/log-based-incremental-replication.md
```diff
@@ -0,0 +1,260 @@
...
+
+      During a replication job, Stitch iterates over each message in sequence, or in the same order that the messages were written to the binary log. Data is extracted for any messages that correspond to tables that have been tracked for replication. This means that all tables using {{ page.table }} are extracted from the source database simultaneously as one seamless process.
+
+      To ensure only new and updated data is selected for replication, Stitch will use a binary log's Log Sequence Number (LSN) to 'bookmark' its place in a log file. This value is used to resume replicating data where a previous replication job ended.
+
+    subsections:
+      - title: "Initial replication jobs"
+        anchor: "initial-replication-jobs"
+        content: |
+          During the initial replication job for a table using {{ page.title }}, Stitch will use a `SELECT`-based approach to replicate the table in full. If this were a SQL query, it would look like this:
```
dylan-stitch: should we clarify that we store the MAX LSN that we will return to for log-based replication prior to executing this query?

_replication/replication-methods/log-based-incremental-replication.md
```diff
@@ -0,0 +1,260 @@
...
+        content: |
+          After the historical replication of a table is complete, Stitch will read updates for the table from the database's binary logs. During ongoing replication jobs using {{ page.title }}, a few things will happen:
+
+          1. Stitch reads log messages in the binary file and persists data for tables set to replicate.
+          2. At the end of the replication job, Stitch bookmarks its place in the log file by storing its current LSN.
+          3. During the **next** replication job, Stitch will resume reading data with a **greater** LSN than the LSN from the previous job.
+          4. At the end of the replication job, Stitch bookmarks its place in the log file again.
+          5. Repeat.
+
+          While {{ page.title }} isn't a SQL-based approach, the following SQL query demonstrates how ongoing replication jobs using {{ page.title }} would look as SQL:
```
dylan-stitch: I honestly think that this will confuse people

_replication/replication-methods/log-based-incremental-replication.md
```diff
@@ -0,0 +1,260 @@
...
+
+      - title: "Limitation 5: Frequent updates can increase row usage"
+        anchor: "limitation-5--row-usage"
+        content: |
+          While {{ page.title }} is the most accurate way to replicate your data, it has the potential to consume a large amount of your row quota. To factors can contribute to this:
+
+          1. Each modification to a record may a single log message for the modification, and
+          2. Stitch reads log messages in order during extraction
+
+          This means that each log message Stitch extracts counts as a replicated row, even if all the messages are for the same record.
```
dylan-stitch: if multiple updates are loaded within the same batch, they should not actually be counted as multiple records, though this would add a level of complexity to the explanation that I'm not sure is worth getting into?

_replication/replication-methods/log-based-incremental-replication.md
erinkcochran87: I wondered about this as well. We can scrap it.

_replication/replication-methods/log-based-incremental-replication.md
erinkcochran87: I'd thought they *would* be counted towards usage, which is why the section is in here. Is that not the case? If it isn't, and this sort of scenario won't affect the total number of rows, I think this can come out.

_replication/replication-methods/log-based-incremental-replication.md
erinkcochran87: Yeah, I think so, since (I think) that'll be where the second job starts, right? I think I need some clarification on how we determine the max LSN here.

_replication/replication-methods/log-based-incremental-replication.md
```diff
@@ -0,0 +1,260 @@
...
+
+      1. [How it works (with examples)](#how-log-based-incremental-replication-works),
+      2. [When it should be used](#when-log-based-incremental-replication),
+      3. [Limitations of this Replication Method](#limitations), and
+      4. [How to enable it for your database integration](#enabling-log-based-replication)
+
+  - title: "How {{ page.title }} works"
+    anchor: "how-log-based-incremental-replication-works"
+    content: |
+      There are two types of binary log replication: Statement and row-based. Stitch uses a row-based approach, which means that when rows are modified via [a supported event type](#limitation-2--database-event-types), the entire row is written to the binary log file as a log message.
```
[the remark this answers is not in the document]
erinkcochran87: I think this opens up a lot of questions about how and why schema violations occur, then. D'you have any insight on why we wouldn't (or can't) use ordinal info to determine correct field loading, if it's available?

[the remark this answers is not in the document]
dylan-stitch: I talked with Paul and apparently I misunderstood something he said previously. Ignore this one.

_replication/replication-methods/log-based-incremental-replication.md
dylan-stitch: Right. I'm not sure what we do to retrieve the max lsn though

PR #227, opened 2019-01-09
Connect: /v4 destinations

erinkcochran87:
This PR does a lot of things:

- Adds documentation for the new version of the Destinations endpoint (v4), including a new `DELETE` route
- Adds documentation for the new `DELETE` method for `/v4/sources`
- Adds documentation for the new Destination Types endpoint
- Adds Azure SQL Data Warehouse (`azure_sqldw`) as a destination option. **Note**: This isn't the same as Azure databases - Data Warehouse is a specific product type.
- Adds logic to the endpoint templates to check for versioning and version status
- Re-vamps handling of endpoint/method-specific errors. This was done to account for errors potentially changing from version to version.


erinkcochran87:
**Docs about docs changes**:

- **Endpoints**
   - `not-available-until-version`
- **Database form properties**
   - `uses-feature-fields`
- **General**
   - Errors are now kept in data files (_`data/connect/response-codes`). Every object type gets its own file.
- **Source & destination objects**
   - `latest-version`
   - `versions` - List of the available versions for the object and whether they're deprecated
   - `available-methods` - List of available methods for the object; includes the ID, title, method, deprecation info, and short description. (See `destination-object` for a full example).

PR #241, opened 2019-01-22
Add files from first pass at jira/v1 docs

luandy64:


luandy64:
The `changelogs` schema looks right if I compare it to the schema file in `tap-jira`

PR #247, opened 2019-01-30
Fixes to the Choosing a Destination Guide

erinkcochran87:
This PR fixes some display issues in the Choosing a Destination guide that were introduced as a result of #207 .


erinkcochran87:
**Note**: The `docs-about-docs` changes here relate to the addition of `fully-managed`, `pricing-model`, and `release-status` properties in the Destination Reference file templates.

PR #257, opened 2019-02-04
update based on tap behavior change

matt-zach-s:
See code comment here. https://github.com/singer-io/tap-zendesk/blob/0f919fb525835f2fa5faffa7b03e26fa8b089484/tap_zendesk/streams.py#L306

And Nick's mention this is not incremental on JIRA card SRCE-365.


erinkcochran87:
Closing; we determined this morning that while the tap code is confusing, it's correct and this stream is actually `incremental`.

PR #261, opened 2019-02-05
full-table-update

matt-zach-s:
From looking both at the schema for this stream, and the code for this stream I do not believe this is incrementally replicated: https://github.com/singer-io/tap-campaign-monitor/blob/master/tap_campaign_monitor/schemas/campaign_recipients.json


erinkcochran87:
@matt-zach-s I think this actually requires looping in a developer. `campaign_recipients` is a substream of `campaigns`, which is *supposedly* Incremental. If this is indeed the case, I would expect `campaign_recipients` to be incremental based on the rep key `campaigns` uses.

matt-zach-s:
Interesting, that sounds correct to me too.

erinkcochran87:
Closing.

PR #273, opened 2019-02-28
SSH re-vamp

erinkcochran87:
This PR revamps the docs for SSH guides. Because the steps vary by hosting type (ex: Amazon vs. Azure), it was getting difficult to maintain a bunch of includes. It was also difficult for users to figure out which instructions they actually needed to use.

Each hosting type that Stitch supports SSH for (self, Amazon, and Azure, at the time of publishing) now has a dedicated guide. This eliminates redundancy in the docs and hopefully makes searching for these topics a bit easier.


erinkcochran87:
@cmerrick - This is the [direct link](https://github.com/stitchdata/docs/pull/273/files#diff-dd6cb05b8d2ae734eabe53e0672654ab) to the file I asked you to look over.

_account-security/security/encryption/encryption-general.md
```diff
@@ -0,0 +1,113 @@
...
+
+  - name: "Microsoft Azure"
+    guide: "ssh-microsoft-azure"
+    description: "Stitch currently supports connecting Microsoft Azure SQL Server (as an integration) and Azure SQL Data Warehouse (as a destination). Other Microsoft Azure offerings aren't currently supported."
+
+sections:
+  - title: "SSL connections"
+    anchor: "ssl-connections"
+    content: |
+      [Secure Sockets Layer](https://www.verisign.com/en_US/website-presence/online/ssl-certificates/index.xhtml){:target="new"}, or SSL, is a standard security technology used to establish encrypted communication between a web server and a browser. SSL ensures that communication to and from Stitch remains private and secure.
```
cmerrick: We should always refer to TLS alongside SSL. TLS is the successor to SSL, and in reality it is much more common these days.

_account-security/security/encryption/encryption-general.md
```diff
@@ -0,0 +1,113 @@
...
+
+    subsections:
+      - title: "Stitch application access"
```
cmerrick: I think this section is potentially confusing since it talks about SSL in a completely different context. Doesn't the general security FAQ section already cover this topic?

_account-security/security/encryption/encryption-general.md
```diff
@@ -0,0 +1,113 @@
...
+
+          This is also applicable to Stitch's [Amazon Redshift]({{ link.destinations.overviews.redshift | prepend: site.baseurl }}), [Google BigQuery]({{ link.destinations.overviews.bigquery | prepend: site.baseurl }}), [Microsoft Azure SQL Data Warehouse]({{ link.destinations.overviews.azure | prepend: site.baseurl }}), and [Snowflake]({{ link.destinations.overviews.snowflake | prepend: site.baseurl }}) destination offerings.
+
+          Connections to these integrations and destinations will attempt to use verified SSL with no action required on your part.
+
+      - title: "Connections with configurable SSL options"
+        anchor: "connections-configurable-ssl"
+        content: |
+          For non-web-hosted database connections, Stitch may support configurable SSL. To use SSL with a database Stitch supports, the database support and allow SSL connections.
```
cmerrick: "non-web-hosted databases" is a confusing phrase. I think it's a little hard to draw a categorical line around the databases that we handle SSL for automatically versus those that we offer manual configuration for - I recommend just saying,  "For some databases..."

Also - missing a word or two at the end of this sentence, "the database support and allow..." should be something like, "the database must be configured to support and allow..."

_account-security/security/encryption/encryption-general.md
```diff
@@ -0,0 +1,113 @@
...
+      <p>Refer to the <a href="{{ link.security[item.guide] | prepend: site.baseurl }}">SSH tunnels for {{ item.name | append: " databases"}}</a> guide.</p>
+      </td>
+      </tr>
+      {% endfor %}
+      </table> 
+
+  - title: "Advanced connectivity"
+    anchor: "advanced-connectivity"
+    content: |
+      Additional connection options may be implemented as part of a Stitch Enterprise plan. This includes:
```
cmerrick: suggest changing "may be implemented" to "are available"

_account-security/security/encryption/ssh-for-amazon-web-services.md
```diff
@@ -0,0 +1,231 @@
...
+          1. Navigate to the VPC Management Console in AWS. If you use the **Services** menu (top left corner), click the **VPC** option under the **Networking & Content Delivery** section.
+          2. On the VPC Dashboard, click the **Launch EC2 Instances** button.
+          3. Next, you’ll be asked to select the Amazon Machine Image, or AMI, that will be used to launch the instance.
+
+             We’ll be using a Linux-based AMI (like Ubuntu) for this tutorial:
+
+             ![Ubuntu Amazon Machine Image option in AWS]({{ site.baseurl }}/images/shared/ssh/amazon-ec2-ubuntu-ami.png)
+
+             Click the **Select** button next to the AMI you want to use.
+          4. On the next page, you'll select the instance type. Generally, a small instance will work just fine. For example: `t2.medium`. You can find more info about instance types on [Amazon's website](https://aws.amazon.com/ec2/instance-types/){:target="new"}.
```
cmerrick: since "small" is also an instance size, recommend changing to "Generally, a small instance will work just fine. For example: `t2.small`.

_account-security/security/encryption/ssh-for-amazon-web-services.md
```diff
@@ -0,0 +1,231 @@
...
+
+          Review the instance's settings, paying particular attention to the fields highlighted in the image below:
+
+          ![Reviewing the EC2 Instance Details]({{ site.baseurl }}/images/shared/ssh/amazon-ec2-instance-review.png)
+
+          - The **Security Groups** section should list either a new Security Group for Stitch OR an existing group that contains group rules for Stitch's IP addresses. If it doesn't, [refer to Step 2.2](#ec2-instance-security-group) for instructions.
+          - In the **Instance Details** section:
+             - **Network**: This field should contain the ID of the same VPC that the database is in.
+             - **Subnet**: This field should contain the ID of a public subnet.
+             - **Assign Public IP & Assign IPv6 IP:** We strongly recommend using a [public subnet with the instance](#configure-ec2-instance) and auto-assigning a public IP address. This will ensure that Stitch can access the instance.
```
cmerrick: this isn't just a strong recommendation, I think it's a requirement

_account-security/security/encryption/ssh-for-amazon-web-services.md
```diff
@@ -0,0 +1,231 @@
...
+  - title: "Enable the SSH server to access the database"
+    anchor: "enable-ssh-server-access"
+    content: |
+      After the EC2 instance has finished initializing, you can move onto configuring the access rules for database. In this section, you'll create a VPC Security Group that will forward traffic from the SSH server (EC2 instance) to the database in the private subnet.
+    
+    substeps:
+      - title: "Retrieve the VPC's IPv4 CIDR"
+        anchor: "retrieve-vpc-ip"
+        content: |
+          In this step, you'll retrieve the SSH server's IP address, or IPv4 CIDR. This value will be followed by a slash and a number between 0 and 32. For example: `10.0.0.0/16`
```
cmerrick: these instructions actually have the user whitelist the CIDR for the *entire VPC*. This will include the SSH tunnel instance, so it will work, but it might be more permissive than they want

_account-security/security/encryption/ssh-for-amazon-web-services.md
```diff
@@ -0,0 +1,231 @@
...
+
+          - The **Security Groups** section should list either a new Security Group for Stitch OR an existing group that contains group rules for Stitch's IP addresses. If it doesn't, [refer to Step 2.2](#ec2-instance-security-group) for instructions.
+          - In the **Instance Details** section:
+             - **Network**: This field should contain the ID of the same VPC that the database is in.
+             - **Subnet**: This field should contain the ID of a public subnet.
+             - **Assign Public IP & Assign IPv6 IP:** We strongly recommend using a [public subnet with the instance](#configure-ec2-instance) and auto-assigning a public IP address. This will ensure that Stitch can access the instance.
+
+          After you're reviewed the instance's settings, click the **Launch** button in the lower right corner to launch the instance.
+
+          **Note**: It may take a few minutes for the instance creation process to complete. The status in the VPC Dashboard page will change to `Available` when the instance is ready.
```
cmerrick: I don't see the part of the instructions to copy down the instance's public IP to be pasted into Stitch. Am I missing those?

Also - without an elastic IP, the instance's public IP could change... so we probably want to include a step to add an elastic IP.

_account-security/security/encryption/ssh-for-microsoft-azure.md
```diff
@@ -0,0 +1,136 @@
...
+               ```
+             - **Protocol**: Select **TCP**.
+             - **Action**: Select **Allow**.
+             - **Name**: Enter `stitch`.
+          6. When finished, click **Add** to create the inbound rule.
+
+  - title: "Enable the SSH server to access the database"
+    anchor: "enable-ssh-server-access"
+    content: |
+      Next, you'll configure the {{ destination.display_name }} database to allow traffic forwarded from the virtual machine to access the database server. This is accomplished by whitelisting the virtual machine's public IP address in the server's firewall settings.
```
cmerrick: If you use the *public IP* then this tunnel isn't really accomplishing anything. This should be private IP.

_account-security/security/encryption/ssh-for-amazon-web-services.md
erinkcochran87: Agreed. Thanks for pointing it out; it slipped by me.

_account-security/security/encryption/encryption-general.md
erinkcochran87: Change 'login credentials' to something like 'communictation'

_account-security/security/encryption/ssh-for-amazon-web-services.md
erinkcochran87: 1. Get Private IP from the bastion
2. Add `/32` to the IP

_account-security/security/encryption/ssh-for-amazon-web-services.md
erinkcochran87: 1. EC2 > Elastic IPs
2. Follow prompts - make sure to select the bastion as the instance
3. Click the Private IP
4. Click Associate to create the IP

Call out that instances stopping can potentially change the IP address. Users can add an elastic IP to protect against this.

PR #276, opened 2019-03-06
Shiphero/v1

erinkcochran87:
This PR adds documentation for the new ShipHero integration.


erinkcochran87:
**Docs about docs**: This PR adds a new Frontmatter parameter named `use-default-icon` to integration pages. If an integration doesn't have an icon ready in Contentful by the time of doc publication, integrations can now use a placeholder. This way, no broken images are created.

erinkcochran87:
[Here's the preview](https://deploy-preview-276--stitchdocs.netlify.com/docs/integrations/saas/shiphero).

_integration-schemas/shiphero/v1/products.md
```diff
@@ -0,0 +1,296 @@
...
+description: |
+  The `{{ table.name }}` table contains info about
+
+replication-method: "Key-based Incremental"
+api-method:
+    name: "Get products"
+    doc-link: "https://shipheropublic.docs.apiary.io/#reference/products/get-product/get-product"
+
+replication-key:
+  name: "from"
```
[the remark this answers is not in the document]
luandy64: We added [this](https://github.com/singer-io/tap-shiphero/commit/42375165d69c23e0f7a80f0d95a0fbc6e403bc21) to the tap to reflect that

_integration-schemas/shiphero/v1/orders.md
```diff
@@ -0,0 +1,445 @@
...
+description: |
+  The `{{ table.name }}` table contains info about the orders in your {{ integration.display_name }} account.
+
+replication-method: "Key-based Incremental"
+api-method:
+    name: "Get orders"
+    doc-link: "https://shipheropublic.docs.apiary.io/#reference/orders/get-orders/get-orders"
+
+replication-key:
+  name: "from"
```
luandy64: There is no `from` field on the record, but there is a date that we look at when we sync the `orders`.

_integration-schemas/shiphero/v1/shipments.md
```diff
@@ -0,0 +1,424 @@
...
+description: |
+  The `{{ table.name }}` table contains info about the shipments in your {{ integration.display_name }} account.
+
+replication-method: "Key-based Incremental"
+api-method:
+    name: "Get shipments"
+    doc-link: "https://shipheropublic.docs.apiary.io/#reference/shipment/get-shipments/get-shipments"
+
+replication-key:
+  name: "from"
```
luandy64: Same as the `orders` comment.

_integration-schemas/shiphero/v1/orders.md
erinkcochran87: Right - this property shows the rep key as something that isn't an attribute on the record itself. Like a path parameter. I assumed it was called `from` after looking at the API docs, but if we're using something else, I'd like to note it so support understands what they're looking at. @luandy64

erinkcochran87:
Re: Pipedrive: Yes; I intended it to be there.

I'll address the other stuff now.

_integration-schemas/shiphero/v1/orders.md
luandy64: I don't know what we should call it. This stream just runs until it is caught up with "today".

luandy64: Digging a bit deeper, these are the parameters I think: `updated_from` and `updated_to`

https://github.com/singer-io/tap-shiphero/blob/e25ea8a4b4b0b5d93ae33ef42ed1ce1ab544fa39/tap_shiphero/sync.py#L219-L220

erinkcochran87: Hey, that works! Thank you!

PR #281, opened 2019-03-15
Update docs for new historical sync strategy

cmerrick:
Changes: 
 - reflect that historical syncs with a PK will always try to bookmark against that PK
 - encourage users to wait until syncs are complete to analyze data discrepancies

### DIFF
--- a/_replication/replication-methods/log-based-incremental-replication.md
+++ b/_replication/replication-methods/log-based-incremental-replication.md
@@ -94,7 +94,7 @@ sections:
               ORDER BY id
              ```
              
-             Note that Stitch will automatically omit the WHERE clause in this statement if the table has no primary key or a primary key that is not sortable. 
+             **Note**: Stitch will automatically omit the `WHERE` clause in this statement if the table has no Primary Key or a Primary Key that is not sortable. 
 
       - title: "Ongoing replication jobs"
         anchor: "ongoing-replication-jobs"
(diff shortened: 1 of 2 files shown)


_replication/replication-methods/log-based-incremental-replication.md
```diff
@@ -85,10 +85,13 @@ sections:
...
 
              ```sql
-             SELECT column_you_selected_1,
+             SELECT id,
+                    column_you_selected_1,
                     column_you_selected_2,
                     [...]
                FROM schema.table
+              WHERE id < [max id value]
+              ORDER BY id
```
erinkcochran87: @cmerrick I don't think it's entirely obvious what happens if a historical sync spans multiple jobs. Is this accurate?:

1. Historical sync starts; max PK for the table is stored
2. Historical sync doesn't complete during job; bookmark for table is stored
3. Next job starts, historical sync resumes using bookmark
4. Jobs continue replicating historical data using bookmark until complete
5. When historical is complete, switch to logs

I also thought this would only be used if the PK met certain requirements - if that's the case, we should include it here.

cmerrick: I added a note about omitting the WHERE clause when the PK doesn't meet requirements. Your description of the multi-job scenario is accurate, but IMO is too complicated and too much of an edge case to include in these docs.

PR #311, opened 2019-05-21
Oracle/v1: Add support for Key-based Incremental Replication

erinkcochran87:
This PR updates the docs to reflect the newly added support for Key-based Incremental Replication in Oracle integrations.


_data/taps/extraction/replication-methods/key-based-incremental.yml
```diff
@@ -29,7 +29,7 @@ structural-changes: |
   Detected and automatically handled by Stitch
 
 configuration-requirements: |
-  A `timestamp`, `date-time`, or `integer` column suitable as a [Replication Key]({{ link.replication.rep-keys | prepend: site.baseurl }}).
+  A `TIMESTAMP`, `DATETIME`, or `INTEGER` column suitable as a [Replication Key]({{ link.replication.rep-keys | prepend: site.baseurl }}).
```
nick-mccoy: Oracle uses slightly different sql datatypes and I think this should be `DATE`, `TIMESTAMP`, or `NUMBER`

There are a few different types of `TIMESTAMP`s (`TIMESTAMP WITH TIME ZONE`, `TIMESTAMP WITH LOCAL TIME ZONE`), but they are technically sub-types of `TIMESTAMP`

erinkcochran87: This is a general file that refers to all integrations, not Oracle specifically. I can add `NUMBER` to this.

nick-mccoy: Ok, `NUMBER` does cover `INTEGER`, so if this is general there is no need to change it.  Thanks for the clarification!

PR #331, opened 2019-06-27
LivePerson/v1

dmosorast:
This PR adds documentation for the new LivePerson integration.


dmosorast:
@erinkcochran87 From looking at the code, I don't think the replication keys are correct. Since this is a tap that uses tap-framework it's a bit hard to follow, but it looks like the method for updating the state is using the `incorporate` function in `state.py`.

Here are all of the usages I can find with github search: https://github.com/fishtown-analytics/tap-liveperson/search?q=incorporate&unscoped_q=incorporate

From that, it looks like only three streams use state:

- `engagement_history` and `messaging_interactions` [use `BaseStream`'s call](https://github.com/fishtown-analytics/tap-liveperson/search?q=BaseStream&unscoped_q=BaseStream) to incorporate, which takes `updated_at` out of the record to store.
- `agent_activity` [uses `BaseFileStream`'s call](https://github.com/fishtown-analytics/tap-liveperson/search?q=BaseFileStream&unscoped_q=BaseFileStream) to incorporate, which bookmarks on (what looks like) the beginning of the 7-day window being requested.

PR #364, opened 2019-08-27
Update SF blacklisted objects

dylan-stitch:
added new objects based on https://github.com/singer-io/tap-salesforce/blob/master/tap_salesforce/salesforce/__init__.py


erinkcochran87:
And updates made in https://github.com/singer-io/tap-salesforce/pull/68 and https://github.com/singer-io/tap-salesforce/pull/64

PR #367, opened 2019-09-05
Opt-out replication error notifications

erinkcochran87:
This PR updates the docs for the new opt-out of replication error email notifications feature.


mwfrantz:
Opt out launched, and now so have the docs.

PR #372, opened 2019-09-25
Klaviyo/v1

erinkcochran87:
This PR adds documentation for the Klaviyo integration.


_integration-schemas/klaviyo/lists.md
```diff
@@ -0,0 +1,44 @@
...
+name: "lists"
+doc-link: ""
+singer-schema: "https://github.com/singer-io/tap-klaviyo/blob/master/tap_klaviyo/schemas/lists.json"
+description: ""
+
+replication-method: "Full Table"
+
+api-method:
+    name: ""
+    doc-link: "https://www.klaviyo.com/docs/api/metrics"
```
erinkcochran87: This is a little tricky. The tap repo says it uses the Metrics API (which it does), but not for this endpoint.

If you look [here](https://github.com/singer-io/tap-klaviyo/blob/master/tap_klaviyo/__init__.py#L12), you'll see the endpoint we're using to get the data for the `lists` table:

```python
ENDPOINTS = {
    'global_exclusions': 'https://a.klaviyo.com/api/v1/people/exclusions',
    'lists': 'https://a.klaviyo.com/api/v1/lists',
    'metrics': 'https://a.klaviyo.com/api/v1/metrics',
    'metric': 'https://a.klaviyo.com/api/v1/metric/',
}
```

And in Klaviyo's docs, there's actually a Lists API in the sidenav. Opening that up, we can see in the list of endpoints that there's a [URL that matches the one we're using](https://www.klaviyo.com/docs/api/lists#lists): `https://a.klaviyo.com/api/v1/lists`

That would make this section look like this:

```
api-method:
  name: "Lists in Account"
  doc-link: "https://www.klaviyo.com/docs/api/lists#lists"
```

_integration-schemas/klaviyo/lists.md
```diff
@@ -0,0 +1,44 @@
...
+
+attributes:
+  - name: "created"
+    type: "string"
+    description: ""
+  
+  - name: "id"
```
erinkcochran87: This should have a `primary-key: true` attribute:

```suggestion
  - name: "id"
    type: "string"
    primary-key: true
    description: ""
```

It also needs to be first in the `attributes` list, so bump this up ahead of `created`.

_integration-schemas/klaviyo/lists.md
```diff
@@ -0,0 +1,44 @@
+---
+tap: "klaviyo"
+version: "1.0"
+
+name: "lists"
+doc-link: ""
+singer-schema: "https://github.com/singer-io/tap-klaviyo/blob/master/tap_klaviyo/schemas/lists.json"
+description: ""
```
erinkcochran87: This needs a description.

```suggestion
description: |
  The {{ table.name }} table contains info about the lists and segments in your {{ integration.display_name }} account.
```

_integration-schemas/klaviyo/global_exclusions.md
```diff
@@ -0,0 +1,32 @@
...
+
+name: "global_exclusions"
+doc-link: ""
+singer-schema: "https://github.com/singer-io/tap-klaviyo/blob/master/tap_klaviyo/schemas/global_exclusions.json"
+description: ""
+
+replication-method: "Full Table"
+
+api-method:
+    name: ""
```
erinkcochran87: This follows the same logic as the comment I left about `api-method` for the `lists` table. To get this data, we're actually using [this endpoint](https://www.klaviyo.com/docs/api/lists#exclusions-list): `https://a.klaviyo.com/api/v1/people/exclusions`

_integration-schemas/klaviyo/global_exclusions.md
```diff
@@ -0,0 +1,32 @@
+---
+tap: "klaviyo"
+version: "1.0"
+
+name: "global_exclusions"
+doc-link: ""
+singer-schema: "https://github.com/singer-io/tap-klaviyo/blob/master/tap_klaviyo/schemas/global_exclusions.json"
+description: ""
```
erinkcochran87: This needs a description.

_integration-schemas/klaviyo/global_exclusions.md
```diff
@@ -0,0 +1,32 @@
...
+
+replication-method: "Full Table"
+
+api-method:
+    name: ""
+    doc-link: "https://www.klaviyo.com/docs/api/metrics"
+
+attributes:
+  - name: "email"
```
erinkcochran87: This should be the table's primary key. I'm getting that from [here](https://github.com/singer-io/tap-klaviyo/blob/master/tap_klaviyo/__init__.py#L67), which isn't a typical implementation, but it's the closest thing I can find.

_integration-schemas/klaviyo/bounce.md
```diff
@@ -0,0 +1,60 @@
...
+api-method:
+    name: ""
+    doc-link: "https://www.klaviyo.com/docs/api/metrics"
+
+attributes:
+  - name: "id"
+    type: "string"
+    primary-key: true
+    description: ""
+    foreign-key-id: "bounce-id"
```
erinkcochran87: Let's comment this out for now, since it doesn't point to any other tables:

```suggestion
    # foreign-key-id: "bounce-id"
```

Do this for all the other tables as well.

_connect-files/api/objects/form-properties/sources/saas/klaviyo-object.md
```diff
@@ -0,0 +1,51 @@
...
+
+description: ""
+
+# -------------------------- #
+#      OBJECT ATTRIBUTES     #
+# -------------------------- #
+
+uses-start-date: true
+
+# object-attributes:
```
erinkcochran87: There's actually a property users need to provide to create this source using the API.

PR #395, opened 2019-11-21
darksky v1

jazsmi:
This PR adds documentation for the new Darksky integration.


_integration-schemas/darksky/forecast.md
```diff
@@ -0,0 +1,292 @@
...
+
+api-method:
+  name: "Time Machine Request"
+  doc-link: "https://darksky.net/dev/docs#time-machine-request"
+
+attributes:
+  - name: "forecast_date"
+    type: "date-time"
+    primary-key: true
+    description: ""
```
erinkcochran87: This is an attribute the tap generates to extract data from the API, meaning there's no description of it in Darksky's API docs. There should still be one, though, since it determines how and what data is extracted during a replication job.

The [request parameters for the endpoint that the tap uses](https://darksky.net/dev/docs#time-machine-request) give us this clue:

> Either be a UNIX time (that is, seconds since midnight GMT on 1 Jan 1970) or a string formatted as follows: [YYYY]-[MM]-[DD]T[HH]:[MM]:[SS][timezone]. timezone should either be omitted (to refer to local time for the location being requested), Z (referring to GMT time), or +[HH][MM] or -[HH][MM] for an offset from GMT in hours and minutes. The timezone is only used for determining the time of the request; the response will always be relative to the local time zone.

And the endpoint is: `https://api.darksky.net/forecast/[key]/[latitude],[longitude],[time]`

Think of `[time]` as equivalent to `forecast_time`, as we have it in the table here.

TL:DR; This is the date that a forecast record pertains to for the given `latitude` and `longitude`.

_integration-schemas/darksky/forecast.md
```diff
@@ -0,0 +1,292 @@
+---
+tap: "darksky"
+version: "1.0"
+
+name: "forecast"
+doc-link: "https://darksky.net/dev/docs#overview"
+singer-schema: "https://github.com/singer-io/tap-darksky/blob/master/tap_darksky/schemas/forecast.json"
+description: "This table contains information about the forcasts."
```
erinkcochran87: While this description is technically correct, I think there's probably more we should include here after reviewing Darksky's API docs. 

For example:
1. The tap only retrieves data for the `daily` and `hourly` data blocks, but there are a bunch of other ones we exclude.
2. A single record in the table always pertains to a given date (`forecast_date`), *and*
3. A single record in the table always pertains to a given location (The `locations` the user enters into Stitch)
4. The units that measurements are returned in is determined by what the user enters into the **Units** field in Stitch. I can see folks writing into support about discrepancies if this isn't clear, so let's repeat it a bunch to try and prevent that.

Let's try to make it obvious to the reader what's available before they even open the table schema:

```suggestion
description: |
  The `forecasts` table contains weather conditions for a particular date and location. The locations are determined by the locations entered into the **Locations** field in Stitch.
  
  **Note**: The units data points are returned in is determined by the value entered into the **Units** field in Stitch. For example: If `us` is entered, data will be returned in Imperial units. 
```

_connect-files/api/objects/form-properties/sources/saas/darksky-object.md
```diff
@@ -0,0 +1,71 @@
...
+# Only source-specific attributes need to be listed here.
+# The following attributes are considered common,
+# and therefore don't need to be listed:
+# anchor_time, cron_expression, frequency_in_minutes, image_version, start_date 
+
+object-attributes:
+  - name: "language"
+    type: "string"
+    required: true
+    description: "The language."
```
erinkcochran87: Based on Darksky's API docs, this is what this setting does:

> Return `summary` properties in the desired language.

Meaning that the `summary` attributes in the `forecast` table are returned in this language. I'd make this description more specific, and also include a link to the list of values that the API will accept.

I see that their API docs don't include heading anchors, so maybe something like this:

```suggestion
    description: |
      The language that `summary` attributes in the `forecast` table should be returned in. For example: If set to `en`, `summary` data will be returned in English.
      
      Refer to [{{ form-property.display-name }}'s documentation](https://darksky.net/dev/docs#time-machine-request){:target="new"} for a list of accepted values.
```

_connect-files/api/objects/form-properties/sources/saas/darksky-object.md
```diff
@@ -0,0 +1,71 @@
...
+    value: "en"
+
+  - name: "location_list"
+    type: "string"
+    required: true
+    description: "The longitude and latitude of the location the app was used."
```
erinkcochran87: Like the UI, I'm assuming that a user could provide more than one location for this property. Is that the case, @jazsmi? If so, this description should be adjusted to reflect that and include the format that values have to be in to be accepted. Ex: `<latitude>,<longitude>` for a single location, and `<latitude>,<longitude>;<latitude>,<longitude>; ... etc` for multiple.

_connect-files/api/objects/form-properties/sources/saas/darksky-object.md
```diff
@@ -0,0 +1,71 @@
...
+  - name: "secret_key"
+    type: "string"
+    required: true
+    description: "The secret API Key"
+    value: "YOUR_SECRET_KEY"
+    
+  - name: "units"
+    type: "string"
+    required: true
+    description: "The unit of measurement."
```
erinkcochran87: This description should be expanded to describe what this property actually does.

_integration-schemas/darksky/foreign-keys.md
```diff
@@ -0,0 +1,21 @@
+---
+# -------------------------- #
+#     USING THIS TEMPLATE    #
+# -------------------------- #
+
+## NEED HELP USING THIS TEMPLATE? SEE:
+## https://docs-about-stitch-docs.netlify.com/reference/integration-templates/saas-foreign-keys/
+## FOR INSTRUCTIONS & REFERENCE INFO
+
+tap-reference: "integration"
```
erinkcochran87: I don't think this file is actually needed, since the integration only returns a single table. I'd say delete it.

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+## NEED HELP USING THIS TEMPLATE? SEE:
+## https://docs-about-stitch-docs.netlify.com/reference/integration-templates/saas/
+## FOR INSTRUCTIONS & REFERENCE INFO
+
+
+# -------------------------- #
+#      Page & Formatting     #
+# -------------------------- #
+
+title: Dark Sky
```
erinkcochran87: It looks like this is actually `Darksky`

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+keywords: darksky, integration, schema, etl darksky, darksky etl, darksky schema
+layout: singer
+# input: false
+
+# -------------------------- #
+#         Tap Details        #
+# -------------------------- #
+
+name: "darksky"
+display_name: "Dark Sky"
```
erinkcochran87: It looks like this is actually `Darksky`

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+status: "Open Beta"
+certified: false
+
+historical: "1 year"
+frequency: "30 minutes"
+tier: "Free"
+status-url: "https://status.darksky.net/"
+
+anchor-scheduling: true
+cron-scheduling: false
```
erinkcochran87: This is supported, and so should be `true`. Let me know if you need a refresher on how to figure out the support for this feature.

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+
+## incompatible:
+  ## [redshift]: "always,sometimes,never"
+  ## reason: "copy" 
+
+# -------------------------- #
+#      Setup Instructions    #
+# -------------------------- #
+
+requirements-list:
```
erinkcochran87: These are fields in the UI - not prerequisites to the setup process. Let's move this info down into the `add integration` setup step.

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+  - item: "**Language**. The language will be a two-letter value. Example: 'en' for English, 'es' for Spanish, and 'ar' for Arabic."
+  - item: "**Location List**. The lattitude and longitude of the locations that weather data should be returned for."
+  - item: "**Secret Key**. The secret key grants Stitch access to the API. The key can be created on your Dark Sky API account home page at: https://darksky.net/dev/account"
+  - item: "**Units**. This is the measurement system that the weather data will be returned. Example: 'us' for Imperial Units, and 'si' for International System of Units."
+
+requirements-info: "To find the list of available languages and units, visit the Dark Sky Forecast Request docs at https://darksky.net/dev/docs#forecast-request"
+
+setup-steps:
+  - title: "Get Secret Key"
```
erinkcochran87: I'd make this more specific - like `"Retrieve your {{ integration.display_name }} secret key"`

On another note, only use title case for page titles. Section headings use sentence case.

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+    anchor: "secret-key"
+    content: |
+      1. Log into your Dark Sky API account.
+      2. On your account home page, your Secret Key is available at the top of the page. You will use this Secret Key to add your integration.
+  - title: "add integration"
+    # content: |
```
erinkcochran87: This should be uncommented, and contain the fields in the integration's settings page in the UI. You currently have these things in `requirements-list` - move them to here. 

It'd then be something like this:

```
content: |
  4. In the **Language** field, [instructions for this field]
  5. In the **Location List** field, [instructions for this field]
  6. In the **Secret Key** field, paste your {{ integration.display_na,e }} secret key that you retrieved in [Step 1](#secret-key).
  7. In the **Units** field, [instructions for this field]
```

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+      # starting with 4., add instructions for additional fields in UI
+  - title: "historical sync"
+  - title: "replication frequency"
+  - title: "track data"
```
erinkcochran87: Does the integration support table or column selection? If not (and the [`table-selection` and `column-selection` properties](https://github.com/stitchdata/docs/pull/395/files#diff-a4de8023d1ecde82002504703dddcd23R52) make it seem like it doesn't, since they're both currently `false`) this line shouldn't be here.

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+  - item: "**Secret Key**. The secret key grants Stitch access to the API. The key can be created on your Dark Sky API account home page at: https://darksky.net/dev/account"
+  - item: "**Units**. This is the measurement system that the weather data will be returned. Example: 'us' for Imperial Units, and 'si' for International System of Units."
+
+requirements-info: "To find the list of available languages and units, visit the Dark Sky Forecast Request docs at https://darksky.net/dev/docs#forecast-request"
+
+setup-steps:
+  - title: "Get Secret Key"
+    anchor: "secret-key"
+    content: |
+      1. Log into your Dark Sky API account.
```
erinkcochran87: Use `{{ integration.display_name }}` when including the integration's name anywhere in page content. It makes it easier to update things if/when the name changes.

Ex: `Log into your {{ integration.display_name }} API account.`

In this case, I'd also include a link right to the place where users should log in (https://darksky.net/dev). They may go to `darksky.net` and be confused when they can't log in (not our fault, but it might reduce the number of questions that support gets about it).

_saas-integrations/darksky/v1/darksky-v1.md
```diff
@@ -0,0 +1,111 @@
...
+
+# -------------------------- #
+#       Stitch Details       #
+# -------------------------- #
+
+status: "Open Beta"
+certified: false
+
+historical: "1 year"
+frequency: "30 minutes"
```
erinkcochran87: It looks like the default is actually `1 hour`. This is in the UI in the **Replication Frequency** section, by the `Use integration default` checkbox.

PR #402, opened 2019-12-16
Developer docs/IAPI re-write and dev hub

erinkcochran87:
This PR finishes the IAPI doc re-write and developer hub.


erinkcochran87:
Hey @mdelaurentis - here are the links you'll need for the review: [Preview](https://deploy-preview-402--stitchdocs.netlify.com/docs/developers/import-api/guides/sequencing-data-for-the-import-api) | [File](https://github.com/stitchdata/docs/pull/402/files#diff-e55ae0f5bc8d35986d31da47302c66c5R57)

**Preview* will take you to a rendered preview of the guide, and **File** to the file as it exists in this PR. Let me know if you have any questions!

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
```diff
@@ -0,0 +1,204 @@
...
+      - [{{ subsection.title }}](#{{ subsection.anchor }})
+      {% endfor %}
+
+    subsections:
+      - title: "Sequencing and Replication Keys"
+        anchor: "sequence-replication-keys"
+        content: |
+          Think of `sequence` like a [Replication Key]({{ link.replication.rep-keys | prepend: site.baseurl }}).
+
+          For the Import API, it functions similarly to an `updated_at` property: Only records that have a `sequence` value greater than **[TODO- Confirm that this has to be greater than the saved sequence value for the TABLE and not the record. Pretty sure this is the case, but it should be confirmed.]**
```
mdelaurentis: It has to be greater than the saved sequence number for any existing record with the same primary key. You can see the logic for removing records from the batch that _don't_ have a greater sequence number than the existing record here: https://github.com/stitchdata/loader-snow/blob/master/src/clj/com/stitchdata/loader_snow/models/ddl.clj#L298

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
```diff
@@ -0,0 +1,204 @@
...
+
+    subsections:
+      - title: "Generating sequence values"
+        anchor: "generate-sequence-values"
+        content: |
+          When developing your Import API script, you'll need to define how values for the `sequence` property are generated. 
+
+          A simple solution is just to use the current timestamp, but before doing so, consider the following:
+
+          - **How frequently are the records being updated?** Records that are updated every few milliseconds can result in failure if records with identical key values are pushed simultaneously. This means that records with the same Primary Key values cannot be sent during the same clock resolution.
```
mdelaurentis: I could be wrong, but I doubt this would cause the load to fail. I think it would just result in maybe the wrong version of the record being loaded.

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
```diff
@@ -0,0 +1,204 @@
...
+
+             For example: If the resolution is measured in milliseconds, records with identical Primary Key values cannot be sent during the same millisecond.
+
+          - **Are records being pushed from multiple sources?** If so, the time clocks of these sources must be synchronized. This is especially important if different sources are pushing records to the same table.
+
+      - title: "Tracking a table's maximum sequence"
+        anchor: "track-table-maximum-sequence"
+        content: |
+          In addition to generating sequence values for records, you'll also need to track the maximum sequence value for each table being pushed to the Import API. Sequence values need to continually increase or records in the destination won't be updated with new data.
```
mdelaurentis: I'm not sure that it's really true that you need to track the max sequence value for the table. In fact I doubt most of our own integrations do that. I believe most of our integrations just use either the timestamp in millis or some function of the timestamp in millis. Here's where target-stitch adds the sequence number. It basically uses the timestamp at which it received the batch, times 1,000, as a starting point, and then auto-increments after that. So for example if the current time is 123 milliseconds, it would number the records in the batch 123000, 123001, 123002, etc...

The important thing is that the sequence number for a single primary key increases over time.

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
erinkcochran87: Ah, perhaps that wording should be changed - I think `failure` here just means `This version of a record won't load successfully`, not that the load itself will fail.

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
erinkcochran87: This is a weird one - I've seen support conversations about this that were fairly recent. Any ideas who we should run this by?

dmosorast: Yup, this is the line in `target-stitch` that describes Mike's message. There's no need to track with this method, since it's automatically offset using the time + index of the record inthe the current batch (zero-padded) https://github.com/singer-io/target-stitch/blob/master/target_stitch/__init__.py#L421

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
```diff
@@ -0,0 +1,204 @@
...
+    content: |
+      {% include note.html type="single-line" content="**Note**: All example requests are using the Batch endpoint." %}
+
+    subsections:
+      - title: "TODO"
+        anchor: "todo"
+        content: |
+          This request contains three data points: One for record `id: 1`, and two for `id: 2`.
+
+          If these data points were received in this order, only data points 1 and 2 would continue to Stitch. Data point 3 would not, as its `sequence` value is less than the `sequence` value for data point 2.
```
dmosorast: Hm... is this true? I don't see any code in the gate that checks that the sequence numbers are strictly increasing across the `messages` array, though maybe it's at load time where this would happen, not sure.

mdelaurentis: @dmosorast yeah, that happens in the loader. Note that data point 3 would be omitted because its sequence number is less than the sequence number previously observed for another record _with the same primary key_.

_developer-content/import-api/guides/sequencing-data-for-the-import-api.md
mdelaurentis: @erinkcochran87 I'm pretty sure what I said originally in this thread is true. Although this stuff is tricky and it's actually implemented separately in each loader (though as far as I know they generally strive to have the same behavior). If you send me a link to a support conversation I can try to reconcile that with my understanding of how sequence numbers work.

erinkcochran87: :+1: I can't find the support conversation at the moment, but if I happen to stumble across it at some point, I'll send it your way. I think for now we can take this out, since it may be too technical for general uses anyway.

PR #411, opened 2020-01-21
Google Analytics - Cohort Report support

jazsmi:
This commit was to add to the notes in the "Metrics and Dimensions" section about Stitch not supporting Cohort Reports.

### DIFF
--- a/_saas-integrations/google-analytics/google-analytics-latest.md
+++ b/_saas-integrations/google-analytics/google-analytics-latest.md
@@ -119,7 +119,7 @@ setup-steps:
 
       4. **Segments and Filters aren't currently supported.** If you're interested in us adding these features, please get in touch with us.
 
-      5. **Cohort reports aren't currently supported.** Stitch's {{ integration.display_name }} integration does not have the feature for you to specify any cohorts.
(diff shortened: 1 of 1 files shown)


_saas-integrations/google-analytics/google-analytics-latest.md
```diff
@@ -119,6 +119,8 @@ setup-steps:
 
       4. **Segments and Filters aren't currently supported.** If you're interested in us adding these features, please get in touch with us.
 
+      5. **Cohort reports aren't currently supported.** Stitch's {{ integration.display_name }} integration does not have the feature for you to specify any cohorts.
```
erinkcochran87: This is the perfect spot for this to go! I think we should clarify things here a bit, though.

Cohort _reports_ is a bit misleading - it's a specific set of metrics that our integration doesn't support. It's linked in JIRA, but this is the specific metrics the doc request is referring to: https://developers.google.com/analytics/devguides/reporting/core/dimsmets#view=detail&group=lifetime_value_and_cohorts

It's **Lifetime value and cohort metrics.** I think we should be specific here about what we're not supporting, since users will probably know those items by that name.

So, maybe:

> **Lifetime value and cohort metrics aren't currently supported.** If you're interested in us adding these features, get in touch with us.

PR #417, opened 2020-01-27
Google-sheets/v1

dmosorast:
Initial PR to setup locations for google sheets docs to reside as we move into Beta.


erinkcochran87:
Looks like we didn't catch this before creating our own branch - whoops 🤷‍♀ This is a dupe of #420, closing!

PR #418, opened 2020-01-27
Database Integrations - Limit database user table access

erinkcochran87:
This PR:

- Updates the setup instructions for database integrations that require a database user to recommend only granting access to table the user wants to replicate
- Adds a guide about the issues that might be encountered due to a large number of tables being accessed during discovery


erinkcochran87:
[Preview](https://deploy-preview-418--stitchdocs.netlify.com/docs/troubleshooting/too-many-tables-to-load-database-integration) | [File](https://github.com/stitchdata/docs/pull/418/files#diff-fc8bd6ceee5a8c152feffd0ae942ac7c)

PR #434, opened 2020-02-26
Ga360-v1

jazsmi:
This PR adds documentation for the new Google Analytics 360 integration.


jazsmi:
@erinkcochran87 the last two commits "Update ga360-object.md" removes the extra quotation marks in the object-attributes descriptions

jazsmi:
@erinkcochran87 
- The last two "update ga360-object" commits remove the extra quotation marks that I accidentally left in the object-attributes descriptions.
- Was it safe to assume the schema files were Full Table replications?

_integration-schemas/ga360/ga_sessions.md
```diff
@@ -0,0 +1,280 @@
...
+    description: "This section contains any user-level or session-level custom dimensions that are set for a session. This is a repeated field and has an entry for each dimension that is set."
+    subattributes:
+      - name: "index"
+        type: "number"
+        description: "The index of the custom dimension."
+      - name: "value"
+        type: "string"
+        description: "The value of the custom dimension."
+
+  - name: "date"
```
erinkcochran87: This is the table's replication key - it looks like it is an actual field!

Bump this up after all the primary key attributes and add the `replication-key: true` flag. It's probably a good idea to also include a link to the replication section you're gonna write in the description for this attribute.

_integration-schemas/ga360/ga_sessions.md
```diff
@@ -0,0 +1,280 @@
...
+            description: "Descriptive string for the targeting criterion."
+          - name: "customerId"
+            type: "number"
+            description: "The Google Ads Customer ID."
+          - name: "gclId"
+            type: "string"
+            description: "The Google Click ID."
+          - name: "isVideoAd"
+            type: "boolean"
+            description: "True if it is a Trueview video ad."
```
erinkcochran87:
```suggestion
            description: "`TRUE` if it is a Trueview video ad."
```

jazsmi:
@erinkcochran87 Heads up, I somehow managed to paste an attribute name on line 10 in the foreign keys file.

_saas-integrations/ga360/v1/ga360-v1.md
```diff
@@ -0,0 +1,163 @@
...
+## incompatible:
+  ## [redshift]: "always,sometimes,never"
+  ## reason: "copy" 
+
+# -------------------------- #
+#      Setup Instructions    #
+# -------------------------- #
+
+requirements-list:
+  - item: "**A Google BigQuery and {{ integration.display_name }} account**. You need to have your BigQuery account configured to export to {{ integration.display_name }}. To learn how to configure this export, use [Google's step-by-step instructions](https://support.google.com/analytics/answer/3416092?hl=en&ref_topic=3416089)."
```
dmosorast: I think this is backwards. It should possibly be like this:

> You need to have your {{ integration.display_name }} account configured to export to BigQuery.

_saas-integrations/ga360/v1/ga360-v1.md
```diff
@@ -0,0 +1,163 @@
...
+    substeps:
+      - title: "Define the service account details"
+        anchor: "define-service-account-details"
+        content: |
+          {% include destinations/google-bigquery/create-iam-service-account.html type="define-service-account-details" %}
+
+      - title: "Assign BigQuery Admin permissions"
+        anchor: "assign-bigquery-admin-permissions"
+        content: |
+          {% include destinations/google-bigquery/create-iam-service-account.html type="assign-bq-admin" %}
```
dmosorast: I'm not sure we need admin permissions in this service account since we're only running selects (vs. loader which has to create datasets, etc.) but I don't know what kinds of permissions this was tested with.

_integration-schemas/ga360/ga_session_hits.md
```diff
@@ -0,0 +1,708 @@
...
+tap: "google-analytics-360"
+version: "1"
+key: "session_hits"
+
+name: "ga_session_hits"
+doc-link: "https://support.google.com/analytics/answer/3437719?hl=en&ref_topic=3416089"
+singer-schema: "https://github.com/singer-io/tap-ga360/blob/master/tap_ga360/schemas/ga_session_hits.json"
+description: "This table contains information about Google Analytics session hits."
+
+replication-method: "Key-based Incremental"
```
erinkcochran87: This table is missing a Replication Key.

_saas-integrations/ga360/v1/ga360-v1.md
```diff
@@ -0,0 +1,160 @@
...
+
+# -------------------------- #
+#     Replication Details    #
+# -------------------------- #
+
+replication-sections:  
+  - content: |
+      While Stitch's {{ integration.display_name }} replicates the full tables for your dataset, tables are updated by the `ga_sessions_YYYYMMDD` non-field bookmark value. 
+
+      Daily, Google creates an export file, `ga_sessions_YYYYMMDD`. Intraday data is imported to the table `ga_sessions_intraday_YYYYMMDD` around three times per day. Once the final intraday import is complete, then the `ga_sessions_YYYYMMDD` table is complete and can be exported.
```
dmosorast: This might be misleading because we completely ignore the intraday tables in the tap.

It might be able to be removed, if there's a piece of google documentation that talks about how this works, that might be good for further reading, but as far as replication goes, we don't even look at `intraday`.

_saas-integrations/ga360/v1/ga360-v1.md
```diff
@@ -0,0 +1,160 @@
...
+
+      Stitch's {{ integration.display_name }} will re-replicate your data when Google exports a `ga_sessions_YYYYMMDD` table with the date portion greater than the previous export. Allow 24 hours to see complete updates of your data.
```
dmosorast: I think this should say `replicate` instead of `re-replicate`. We only ever pull a table once.

If the sync gets interrupted while we're pulling a table, the tap will request that whole table again the next time, so some records will be re-replicated, but it's not the regular path.

_saas-integrations/ga360/v1/ga360-v1.md
erinkcochran87: This is my fault @jazsmi - I should've had us talk to Dan first. We can chat about this if you want.

PR #435, opened 2020-03-02
adding tools compatibility table

laurenhallden:
- Adds a partial file that is a matrix of popular Stitch destinations, popular analysis tools, and whether the tool supports a native connection to the destination
- Links to external docs where a native connector is documented
- Adds this partial to a new section of docs/analysis-tools and well as /docs/destinations/choosing-a-stitch-destination


laurenhallden:
@erinkcochran87 we might wanna deal with this in another branch, but on docs/analysis-tools there seem to be some default tooltips appearing in the other tables (not the one I just made). They say TOOLTIP. I couldn't immediately figure out how to disable those.

erinkcochran87:
> @erinkcochran87 we might wanna deal with this in another branch, but on docs/analysis-tools there seem to be some default tooltips appearing in the other tables (not the one I just made). They say TOOLTIP. I couldn't immediately figure out how to disable those.

This is an error on my part. I forgot to add a filter that replaces the `TOOLTIP` copy with something that's actually helpful.

Ex: Right now, in one of the tables, there's this bit that determines which icon should show up:

```
  <td valign="top">
  {% case tool.webapp %}
  {% when true %}
  {{ supported }}
  {% when false %}
  {{ not-supported }}
  {% endcase %}
  </td>
```

But it should look like this:

```
  <td valign="top">
  {% case tool.webapp %}
  {% when true %}
  {{ supported | replace: "TOOLTIP","A web app is available." }}
  {% when false %}
  {{ not-supported | replace: "TOOLTIP","A web app isn't currently available." }}
  {% endcase %}
  </td>
```

erinkcochran87:
@laurenhallden I fixed the tooltips in the other tables on top of making the table dynamic. There are instructions for adding destinations in the `destination-compatibility.html` file; and an explanation of the data it's looking at in the `analysis-integrations.md` file. Everything for the compatibility table is in the `compatibility-list` property in this file.

I hope this makes sense - let me know if it doesn't. If it looks good to you, I'll approve this and merge it in.

PR #439, opened 2020-03-04
SaaSOptics/v1

erinkcochran87:
This PR adds documentation for the new SaaSOptics integration.


_saas-integrations/saasoptics/saasoptics-v1.md
```diff
@@ -0,0 +1,120 @@
...
+  - title: "Obtain your token"
+    anchor: "obtain-token"
+    content: |
+      1. Login to the {{ integration.display_name }} account you want to replicate data from.
+      2. In the upper right hand side of the page, click on the user account button that shows your username.
+      3. Click on **API**.
+
+         ![SaaSOptics API hyperlink.]({{ site.baseurl }}/images/integrations/saasoptics-api-button.png){:style="max-width: 550px;"}
+         {:start="4"}
+      4. Generate an API token and keep note of it - you will use it for the Stitch integration.
```
erinkcochran87: > you will use it for the Stitch integration.

Consider rephrasing - this is what they're aiming to do right now, so it's a bit redundant. Maybe `- you'll need it to complete the next step.` or something similar.

_developer-files/connect/api/objects/form-properties/sources/saas/saasoptics-object.md
```diff
@@ -0,0 +1,70 @@
...
+    type: "string"
+    required: true
+    description: "The subdomain of the SaaSOptics server where the account is running."
+    value: "<YOUR_SUBDOMAIN>"
+    
+  - name: "token"
+    type: "string"
+    required: true
+    description: |
+      The API token. This token allows Stitch to access your {{ form-property.diaplay-name }} account's API. Refer to the [{{ form-property.display-name }} documentation]({{ doc-link | append: "#obtain-token" }}) for instructions on retrieving this credential.
```
erinkcochran87: The formatting looks great here ;)

A few suggestions:

1. `The API token.` > Perhaps `A SaaSOptics API token.`, to make it a bit more specific. In this case, I think it'll help orient the reader to the section they're looking at (SaaSOptics Form Property) due to the page being so long and the section header being so far above the table this is in.

2. This is nitpicky (sorry), but the API token doesn't allow Stitch to access the user's SaaSOptics' account API. It just lets Stitch query for their account's data using the SaaSOptics API. To be honest, I don't even think you need this sentence here - I'd take it out to make the description more concise.

_developer-files/connect/api/objects/form-properties/sources/saas/saasoptics-object.md
```diff
@@ -0,0 +1,70 @@
...
+# The following attributes are considered common,
+# and therefore don't need to be listed:
+# anchor_time, cron_expression, frequency_in_minutes, image_version, start_date 
+
+object-attributes:
+  - name: "account_name"
+    type: "string"
+    required: true
+    description: |
+      The name of your unique SaaSOptics database schema, called an "account." This account is the one in which your subscription financial records are managed.
```
erinkcochran87: I think this description overcomplicates what `account_name` actually is. I see you took this from SaaSOptics' docs, which is good, but for our purposes I think their descriptions are a bit overboard.

Consider following the formula you already used for the `token` property description. That description briefly describes what the thing is (`A SaaSOptics API token.`) and then refers the reader to the setup doc that tells them how to get that info.

_developer-files/connect/api/objects/form-properties/sources/saas/saasoptics-object.md
```diff
@@ -0,0 +1,70 @@
...
+    value: "<YOUR_ACCOUNT_NAME>"
+
+  - name: "server_subdomain"
+    type: "string"
+    required: true
+    description: "The subdomain of the SaaSOptics server where the account is running."
```
erinkcochran87: Same here as the `account_name` comment I left. No need to overcomplicate this - it's just as simple as `"The subdomain of your {{ form-property.display-name }} account."` (I see you took this from SaaSOptics' docs, which is good, but for our purposes I think their descriptions are a bit overboard.)

I *would* add, however, that the user shouldn't include the `.saasoptics.com` portion of their subdomain. If they do, the connection will fail because the tap expects just the first part of the string (ex: `stitch`) and not the entire URL (`stitch.saasoptics.com`).

Also, remember to use `{{ form-property.display-name }}` consistently.

_saas-integrations/saasoptics/saasoptics-v1.md
```diff
@@ -0,0 +1,120 @@
...
+
+# -------------------------- #
+#      Feature Summary       #
+# -------------------------- #
+
+feature-summary: |
+  Stitch's {{ integration.display_name }} integration replicates data using the {{ integration.api | flatify | strip }}. Refer to the [Schema](#schema) section for a list of objects available for replication.
+
+
+# -------------------------- #
```
erinkcochran87: Found a setup requirement in [these SaaSOptics docs](https://saasoptics.zendesk.com/hc/en-us/articles/115013917848-Enablement-Authentication) about API tokens:

> **Only an Administrator can actually generate the first Token for a User, but a User can regenerate the Token if needed.** For an existing User, navigate to the User in SaaSOptics (Admin>User) and select the API menu option when viewing the User. Check the box marked "(Re)Generate Token."

I'll explain it if it's unclear, but this is a good research/learning opportunity. What does this mean?

Two changes are needed here - let's chat about them if you're not sure:

1. There should be a requirements list that outlines this - what would that look like?
2. This will also change the instructions for Step 2 - any idea how?

PR #445, opened 2020-03-12
Shopify - Order Refunds table replication note

jazsmi:
Added a replication section:
- suggests creating a separate integration for the order refunds table since it's queried at every single ID

### DIFF
--- a/_saas-integrations/shopify/v1/shopify-v1.md
+++ b/_saas-integrations/shopify/v1/shopify-v1.md
@@ -89,9 +89,9 @@ replication-sections:
   - title: "Order Refunds table replication"
   - anchor: "order-refunds"
   - content: |
-      The Order Refunds file gets queried at every single order ID in the table. If you have this table selected for replication, the process can potentially be very slow depending on how many refunds exist in the table. With Stich's replication process, this could cause other tables to not be replicated for days to weeks at a time. To ensure timely replications of your other selected tables, consider creating a separate integration for only the Order Refunds table.
+      The `order_refunds` table gets queried at every single order ID in the table. If you have this table selected for replication, the process can potentially be very slow depending on how many refunds exist in your account. With Stitch's replication process, this could cause other tables to not be replicated for days to weeks at a time. To ensure timely replication of your other selected tables, consider creating a separate integration for only the `order_refunds` table.
 
-      Note that depending on how much {{ integration.display_name }} data you have, creating a separate integration for your Order Refunds table may negatively affect your API quota.
+      Note that creating a separate integration for your `order_refunds` table may negatively affect your {{ integration.display_name }} API quota.
       
 
 # -------------------------- #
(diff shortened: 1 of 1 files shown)


_saas-integrations/shopify/v1/shopify-v1.md
```diff
@@ -81,6 +81,16 @@ setup-steps:
...
 
+# -------------------------- #
+#     Replication Details    #
+# -------------------------- #
+
+replication-sections:  
+  - title: "Order Refunds table replication"
+  - anchor: "order-refunds"
+  - content: |
+      The Order Refunds table is a Key-based Incremental file that gets queried for every single order ID. If you have this table selected for replication, the process can potentially be very slow depending on how many refunds exist in the table. With Stich's replication process, this could cause other tables to not be replicated for days to weeks at a time. To ensure timely replications of your other selected tables, consider creating a separate integration for only the Order Refunds table.
```
erinkcochran87: I think the first sentence is a bit jargony - how can you simplify this and make the language more relatable?

This may also have API quota implications, which is worth mentioning.

_saas-integrations/shopify/v1/shopify-v1.md
```diff
@@ -89,7 +89,9 @@ replication-sections:
...
   - content: |
-      The Order Refunds table is a Key-based Incremental file that gets queried for every single order ID. If you have this table selected for replication, the process can potentially be very slow depending on how many refunds exist in the table. With Stich's replication process, this could cause other tables to not be replicated for days to weeks at a time. To ensure timely replications of your other selected tables, consider creating a separate integration for only the Order Refunds table.
+      The Order Refunds file gets queried at every single order ID in the table. If you have this table selected for replication, the process can potentially be very slow depending on how many refunds exist in the table. With Stich's replication process, this could cause other tables to not be replicated for days to weeks at a time. To ensure timely replications of your other selected tables, consider creating a separate integration for only the Order Refunds table.
+
+      Note that depending on how much {{ integration.display_name }} data you have, creating a separate integration for your Order Refunds table may negatively affect your API quota.
```
erinkcochran87: A few nitpicky things, but this reads better to me - nice work!

> The Order Refunds file gets queried at every single order ID in the table.

When referring to a specific table, use the table's name - in this case, it's `order_refunds`. We also put table names in backticks (```) to make it stand out.

I'd take the word `file` out of here entirely. It doesn't mean anything in this context - we're not working with files, but tables - and it can be confusing to the reader.

> If you have this table selected for replication, the process can potentially be very slow depending on how many refunds exist in the table.

I'd rephrase the last part of this sentence. It's how many refunds exist in their *account*, not necessarily the table. We're replicating data from their account and putting it into a table, so the table doesn't technically exist until Stitch comes into play.

> With Stich's replication process,

Typo here.

> To ensure timely replications of your other selected tables, consider creating a separate integration for only the Order Refunds table.

`replications` should be singular here. Also, `order_refunds` since you're referring to the name of the table.

>  Note that depending on how much {{ integration.display_name }} data you have, creating a separate integration for your Order Refunds table may negatively affect your API quota.

I think you can make this a bit more general - is the API quota dependent on the amount of data? Or API calls? Or something else? I'd also make it clear that you're referring to the user's **Shopify** API quota here. Right now it could be read as either Stitch or Shopify's API quota.

PR #454, opened 2020-03-24
Workday RaaS v1

jazsmi:
This PR adds documentation for the new Workday RaaS integration.


_saas-integrations/workday/v1/workday-raas-v1.md
```diff
@@ -0,0 +1,133 @@
...
+
+# setup-name: ""
+
+
+# -------------------------- #
+#      Feature Summary       #
+# -------------------------- #
+
+feature-summary: |
+  Stitch's {{ integration.display_name }} integration replicates data using the {{ integration.api | flatify | strip }}. There are no schema files for this integration.
```
erinkcochran87: > There are no schema files for this integration.

This isn't needed - users also won't know what a `schema file` is. That's our internal term.

Also, what data does the integration replicate? How does Stitch get info about what data to replicate? You should put that info in here and point to the `Schema` section (see my comment there) where you can put additional info.

_saas-integrations/workday/v1/workday-raas-v1.md
```diff
@@ -0,0 +1,133 @@
...
+## incompatible:
+  ## [redshift]: "always,sometimes,never"
+  ## reason: "copy" 
+
+# -------------------------- #
+#      Setup Instructions    #
+# -------------------------- #
+
+setup-steps:
+  - title: "Retrieve your report URL."
```
erinkcochran87: Headers don't have periods in them.

_saas-integrations/workday/v1/workday-raas-v1.md
```diff
@@ -0,0 +1,133 @@
...
+    anchor: "retrieve-report-url"
+    content: |
+      1. Login to your {{ integration.display_name }} account.
+      2. Navigate to your reports.
+      3. Select the report you want to replicate in Stitch and ensure that the **Enable As Web Service** box is checked.
+      4. Copy the report URL and keep it available for the next step.
+
+      {% include note.html type=single-line content="**Note**: If you would like to replicate multiple reports, repeat steps 3 & 4." %}
```
erinkcochran87: I don't know that you need a callout here. Those should be reserved for really important info, otherwise you run the risk of everything being important and callouts no longer being meaningful. You could just have this as a separate sentence after your list of instructions.

Also - no ampersands (`&`) - `and` should be used instead.

_saas-integrations/workday/v1/workday-raas-v1.md
```diff
@@ -0,0 +1,133 @@
...
+
+  - title: "add integration"
+    content: |
+      4. In the **Username** field, enter the username for your {{ integration.display_name }} account.
+      5. In the **Password** field, enter the password for your {{ integration.display_ name }} account.
+      6. In the **Report** field, enter the name of your report.
+      7. In the **Report** field, enter the report URL you obtained in [step 1](#retrieve-report-url).
+
+      {% include note.html type=single-line content="**Note**: If you would like to replicate more reports, click on **Configure another report**." %}
```
erinkcochran87: Same here as my other comment about callouts - I think you can just have this as a separate sentence instead of a callout.

_saas-integrations/workday/v1/workday-raas-v1.md
```diff
@@ -0,0 +1,133 @@
...
+#     Integration Tables     #
+# -------------------------- #
+
+# Looking for the table schemas & info?
+# Each table has a its own .md file in /_integration-schemas/saas-integration
+
+
+# Remove this if you don't need it:
+# schema-sections:
+#  - title: ""
```
erinkcochran87: What data should users expect to see when it's loaded into the destination? How are attributes identified by Stitch and replicated? Users want to know this information and outlining it here can help us shape their expectations.

It deals with the reports in Workday. Start there, and see if you can explain how Stitch will identify data points (attributes) in the report. This will also determine how the data looks in the destination.

_developer-files/connect/api/objects/form-properties/sources/saas/workday-raas-object.md
```diff
@@ -0,0 +1,66 @@
...
+  - name: "password"
+    type: "string"
+    required: true
+    description: "Your password for your {{ form-property.display-name }} account."
+    value: "<YOUR_PASSWORD>"
+
+  - name: "reports"
+    type: "string"
+    required: true
+    description: "Your report URL."
```
erinkcochran87: This feels vague to me. What is it, _exactly_? It's the URL of the report you want to replicate, right?

Also, this seems a bit weird to me - how would a user specify multiple reports using the API? There also isn't an attribute equivalent to `Report Name` here - why? These are things that the sources team should know/answer.

_developer-files/connect/api/objects/form-properties/sources/saas/workday-raas-object.md
```diff
@@ -0,0 +1,66 @@
...
+# -------------------------- #
+
+title: "Workday RaaS Source Form Property"
+api-type: "platform.workday-raas"
+display-name: "Workday RaaS"
+
+source-type: "saas"
+docs-name: "workday-raas" # This should be whatever integration.name is. Ex: LinkedIn Ads is linkedin-ads
+
+property-description: ""
```
erinkcochran87: Remember to take this out if you're not using it!

PR #460, opened 2020-03-31
SurveyMonkey/v1 - New setup instructions

ichensvmk:
Fix surveymonkey doc about getting token and survey id.


ichensvmk:
Already signed CLA before.

PR #489, opened 2020-05-14
Google Sheets - Empty worksheets and header row

jazsmi:
This update includes:
- adding a section to the the replication notes mentioning the "skip worksheet" logic
- adds having a header row as a requirement

### DIFF
--- a/_saas-integrations/google-sheets/v1/google-sheets-v1.md
+++ b/_saas-integrations/google-sheets/v1/google-sheets-v1.md
@@ -98,6 +98,10 @@ feature-summary: |
 #      Setup Instructions    #
 # -------------------------- #
 
+requirements-list:
+  - item: |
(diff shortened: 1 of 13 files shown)


_saas-integrations/google-sheets/v1/google-sheets-v1.md
```diff
@@ -118,7 +122,7 @@ setup-steps:
 #     Replication Details    #
 # -------------------------- #
 
-replication sections:  
+replication-sections:
```
erinkcochran87, line 125: I fixed this - this was why none of the info in this section was showing up. It needs to be `replication-sections`, and if you look at the diff (the left side of the *Files Changed* tab), you'll see it was `replication sections` before.

The template this page uses will always look for `replication-sections`

PR #531, opened 2020-08-03
Zoom/v1

jazsmi:
This PR adds docs for the new Zoom (v1) integration.


_integration-schemas/zoom/v1/meeting_poll_results.md
```diff
@@ -0,0 +1,44 @@
+---
+tap: "zoom"
+version: "1"
+key: ""
+
+name: "meeting_poll_results"
+doc-link: ""
+singer-schema: "https://github.com/singer-io/tap-zoom/blob/master/tap_zoom/schemas/meeting_poll_results.json"
+description: |
+  The `{{ table.name }}` table contains information about the results of polls from your {{ integration.display_name }} meetings.
```
erinkcochran87, line 10: Zoom's API docs have some specifications for how the endpoint powering this table works:

```
- Host user type must be Pro.
- Meeting must be a scheduled meeting. Instant meetings do not have polling features enabled.
```

I'd work this into the table description, or folks might wonder why data isn't available for certain meetings.

_integration-schemas/zoom/v1/meeting_poll_results.md
```diff
@@ -0,0 +1,44 @@
...
+    name: "getPastMeetingPolls"
+    doc-link: "https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/listpastmeetingpolls"
+
+attributes:
+  - name: "email"
+    type: "string"
+    primary-key: true
+    description: |
+      The email used for the {{ integration.display_name }} meeting.
+    foreign-key-id: "meeting-email"
```
erinkcochran87, line 24: Is this actually a foreign key? Keep in mind that just because a column is part of a composite Primary Key, it doesn't mean it's necessarily a foreign key.

For example, sometimes dates are part of a composite primary key - let's say a table has a composite PK of `report_id : report_date`. The `report_date` is a PK column, but it's unlikely that it would point to any records outside of that specific table.

_saas-integrations/zoom/v1/zoom-v1.md
```diff
@@ -29,12 +29,109 @@ name: "zoom"
...
+  ## [redshift]: "always,sometimes,never"
+  ## reason: "copy" 
+
+# -------------------------- #
+#      Setup Instructions    #
+# -------------------------- #
+
+requirements-list:
+  - item: |
+      **{{integration.display_name }} authentication apps**. You'll need one or the other to connect your {{integration.display_name }} account to Stitch:
```
erinkcochran87, line 95: This is super nitpicky, but I still think it's worth pointing out. When writing items in here, consider how the first sentence in this entire section (**[integration] Setup requirements**) should make items in this list sound. Ex:

```
## Zoom setup requirements

To set up Zoom in Stitch, you need:

   - **Zoom authentication apps**.
```

While `Zoom authentication apps` is technically correct for what you're talking about here (which are **apps**), the user really only needs a single authentication app, right? To that end, I would change `Zoom authentication apps` to be more specific and read a bit smoother: `A Zoom authentication app`

_saas-integrations/zoom/v1/zoom-v1.md
```diff
@@ -29,12 +29,109 @@ name: "zoom"
...
+        - **OAuth app**. [Click here)](https://marketplace.zoom.us/docs/guides/build/oauth-app) for more information on OAuth apps. If using a {{ integration.display_name }} OAuth App to connect to Stitch, make sure that the following scopes are included in the app:
```
erinkcochran87, line 96: There's an orphaned parenthesis in the `Click here`

PR #551, opened 2020-09-16
Fix tap-kustomer to be community supported

luandy64:
I stumbled on this while working with a Support person


luandy64:
I wanted to remove the change that added the newline at the end of the file, but that's harder than I thought it would be, so I'm going to leave it.

erinkcochran87 (approved):

PR #562, opened 2020-10-05
inSided - coming soon

jazsmi:
coming soon page requested for an alpha


erinkcochran87 (changes requested):
It looks like there are 2 problems here:

1. The are 2 versions of the same file here
2. They're in the `_integration-schemas` folder instead of `_saas-integrations`

erinkcochran87 (approved):

PR #575, opened 2020-10-21
Microsoft Teams/v1

jazsmi:
This PR adds documentation for the new Microsoft Teams integration.


jazsmi:
@erinkcochran87 
1. I ended up taking out the step I added in the Setup Instructions explaining how to create a Microsoft Azure App. I took it out because none of the information would be used in the UI for the integration setup. What I did instead was, in the one of the requirement items, I linked to the section in the Singer repo with those instructions. (I saved the instructions I wrote just in case they should be in there after all.)
2. There are two channels tables: `channel` and `channels`. The `channel` tables has one extra attribute that `channels` doesn't. Other than that, they're the same. I believe I filled them both out anyway.
3. The `channel_messages` and `channel_message_replies` table were given three options as rep keys. I only marked one of them as the rep key in both files. Was this the right thing to do?
4. The `team_device_usage_report` table's PK is supposed to be `id` but it wasn't there. I added it myself - is that okay?

erinkcochran87:
@jazsmi All good questions! I haven't gone through the docs themselves yet, but here are the answers to your questions. I'll ping ya when I'm done with my initial review.

1. For the setup instructions, you actually don't need to do anything here. If you log into Stitch and check out the UI for this integration, you'll see that it's OAuth - all the user has to do is click **Authorize** and complete the auth flow to connect the integration. They don't need to do anything in Azure if they want to just use this integration in Stitch. [Here's what we're doing in Stitch to do this.](https://docs.microsoft.com/en-us/graph/auth-v2-user?context=graph%2Fapi%2F1.0&view=graph-rest-1.0)

   **However**, if they want to set this integration up via the API, then they'll need to complete the steps for setting up that Azure OAuth app. Do you know why this is?

2. I don't think the `channel` table is actually being used anywhere in the tap. I checked in [streams.py](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py) and can't find it. The UI also doesn't list it as a table. I'd ask a dev if this table should be in there - if not, let's have it removed.

3. This is a weird one. While it's possible that a table can have multiple Replication Keys, it's pretty unusual. Usually you'll see tables having several valid Replication Keys; when a user selects that table, they *should* be prompted to select a key.

   I thought that's what was happening here, based on what's in [streams.py](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py#L325), as there's a `replication_key` and a `valid_replication_keys` property:
   ```
    replication_method = 'INCREMENTAL'
    replication_key = 'last_modified_date_time'
    endpoint = 'teams/{group_id}/channels/{channel_id}/messages/delta'
    valid_replication_keys = [
        'last_modified_date_time', 'created_date_time', 'deleted_date_time'
    ]
   ```

   But the UI doesn't prompt you to select a Replication Key when the table is selected. This is what NetSuite Suite Analytics does for tables that have multiple possible Rep Keys. `campaign_subscription_statuses`) is one such table.

   Given that I wasn't prompted to select a Rep Key for the table when I selected it, I would have then expected that the `last_modified_date_time` attribute would have been automatically selected for me, the same way Primary Keys (in this case `id`) are. That isn't the case here.

   So, to summarize: Based on `streams.py`, the Rep Key (at least for `channel_messages`) **should** be `last_modified_date_time`. It's odd that it isn't automatically selected in the UI, nor is there a prompt to select a Replication Key. I would have a dev clarify this, as I think it might lead to extraction problems - will data be successfully extracted for these tables if the Rep Key column isn't selected in the UI? 

   I'm happy to go over this in Zoom too, as it might be easier to explain by showing you this than reading my drivel :)

4. It actually looks like the README has the incorrect PK for `team_device_usage_report`. Based on [streams.py](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py#L517), it should be `user_principal_name: report_refresh_date`. I'd submit a PR to correct this in the README.

   This is a good opportunity to remember that the README isn't infallible. I'd always check it against `streams.py` (or its equivalent) if such a thing exists, as the code is the source of truth.

_integration-schemas/ms-teams/v1/conversation_posts.md
```diff
@@ -0,0 +1,92 @@
...
+doc-link: "https://docs.microsoft.com/en-us/graph/api/conversationthread-list-posts?view=graph-rest-beta&tabs=http"
+singer-schema: "https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/schemas/conversation_posts.json"
+description: |
+  The `{{ table.name }}` table contains information about the posts within a conversation thread in your Microsoft account.
+
+replication-method: "Key-based Incremental"
+
+api-method:
+    name: "List posts"
+    doc-link: "https://docs.microsoft.com/en-us/graph/api/conversationthread-list-posts?view=graph-rest-beta&tabs=http"
```
erinkcochran87, line 16:
```suggestion
    doc-link: "https://docs.microsoft.com/en-us/graph/api/conversationthread-list-posts?view=graph-rest-1.0&tabs=http"
```

[This table actually uses the V1 API ](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py#L477)

_integration-schemas/ms-teams/v1/conversation_threads.md
```diff
@@ -0,0 +1,55 @@
...
+doc-link: "https://docs.microsoft.com/en-us/graph/api/conversation-list-threads?view=graph-rest-beta&tabs=http"
+singer-schema: "https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/schemas/conversation_threads.json"
+description: |
+  The `{{ table.name }}` table contains information about threads in a group conversation in your Microsoft account.
+
+replication-method: "Key-based Incremental"
+
+api-method:
+    name: "List threads"
+    doc-link: "https://docs.microsoft.com/en-us/graph/api/conversation-list-threads?view=graph-rest-beta&tabs=http"
```
erinkcochran87, line 16:
```suggestion
    doc-link: "https://docs.microsoft.com/en-us/graph/api/group-list-threads?view=graph-rest-1.0&tabs=http"
```

[This table actually uses the V1 API ](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py#L443)

_integration-schemas/ms-teams/v1/conversations.md
```diff
@@ -0,0 +1,47 @@
...
+doc-link: "https://docs.microsoft.com/en-us/graph/api/group-list-conversations?view=graph-rest-beta&tabs=http"
+singer-schema: "https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/schemas/conversations.json"
+description: |
+  The `{{ table.name }}` table contains information about conversations within a group in your Microsoft account.
+
+replication-method: "Key-based Incremental"
+
+api-method:
+    name: "List conversations"
+    doc-link: "https://docs.microsoft.com/en-us/graph/api/group-list-conversations?view=graph-rest-beta&tabs=http"
```
erinkcochran87, line 16:
```suggestion
    doc-link: "https://docs.microsoft.com/en-us/graph/api/group-list-conversations?view=graph-rest-1.0&tabs=http"
```

[This table actually uses the V1 API](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py#L415)

_integration-schemas/ms-teams/v1/team_device_usage_report.md
```diff
@@ -0,0 +1,65 @@
...
+    name: |
+      reportRoot: getTeamsDeviceUsageUserDetail
+    doc-link: "https://docs.microsoft.com/en-us/graph/api/reportroot-getteamsdeviceusageuserdetail?view=graph-rest-beta"
+    
+attributes:
+  - name: "id"
+    type: "string"
+    primary-key: true
+    description: "The device ID."
+    #foreign-key-id: "device-id"
```
erinkcochran87, line 24: See my answer to your questions - this table actually has two Primary Keys, neither of which is `id` :)

_integration-schemas/ms-teams/v1/team_drives.md
```diff
@@ -0,0 +1,88 @@
...
+version: "1"
+key: ""
+
+name: "team_drives"
+doc-link: "https://docs.microsoft.com/en-us/graph/api/drive-get?view=graph-rest-beta&tabs=http#get-the-document-library-associated-with-a-group"
+singer-schema: "https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/schemas/team_drives.json"
+description: |
+  The `{{ table.name }}` table contains information about FILL IN THE BLANK in your Microsoft account.
+  
+replication-method: ""
```
erinkcochran87, line 12: This needs a Replication Method.

_integration-schemas/ms-teams/v1/team_drives.md
```diff
@@ -0,0 +1,88 @@
+---
+tap: "ms-teams"
+version: "1"
+key: ""
+
+name: "team_drives"
+doc-link: "https://docs.microsoft.com/en-us/graph/api/drive-get?view=graph-rest-beta&tabs=http#get-the-document-library-associated-with-a-group"
+singer-schema: "https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/schemas/team_drives.json"
+description: |
+  The `{{ table.name }}` table contains information about FILL IN THE BLANK in your Microsoft account.
```
erinkcochran87, line 10: Got a placeholder here

_integration-schemas/ms-teams/v1/users.md
```diff
@@ -0,0 +1,56 @@
...
+doc-link: "hhttps://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-beta"
+singer-schema: "https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/schemas/users.json"
+description: |
+  The `{{ table.name }}` table contains information about users in your Microsoft account.
+
+replication-method: "Full Table"
+
+api-method:
+    name: "List users"
+    doc-link: "https://docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-beta&tabs=http"
```
erinkcochran87, line 16:
```suggestion
    doc-link: "https://docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&tabs=http"
```

[This endpoint actually uses the V1 API](https://github.com/singer-io/tap-ms-teams/blob/master/tap_ms_teams/streams.py#L121)

_data/taps/versions/ms-teams.yml
```diff
@@ -0,0 +1,22 @@
...
+# -------------------------- #
+#     MS TEAMS VERSIONS      #
+# -------------------------- #
+
+latest-version: "1"
+
+released-versions:
+  - number: "1"
+    status: "beta" ## beta, released, deprecated
+    date-released: ""
```
erinkcochran87, line 19: Nothing to do here yet - just leaving a note so we don't forget to fill this in when the beta is ready to go

_developer-files/connect/api/objects/form-properties/sources/saas/ms-teams-object.md
```diff
@@ -0,0 +1,64 @@
...
+
+oauth-description: ""
+
+oauth-attributes:
+  - name: "client_id"
+    type: "string"
+    required: true
+    credential: false
+    description: |
+      Your OAuth application's client ID, obtained when you create an app with Azure.
```
erinkcochran87, line 47: **This** is where I'd link to the instructions in the Singer repo for setting up an Azure app.

_developer-files/connect/api/objects/form-properties/sources/saas/ms-teams-object.md
```diff
@@ -0,0 +1,64 @@
...
+      Your OAuth application's client secret, obtained when you create an app with Azure.
+    value: "<YOUR_CLIENT_SECRET>"
+
+  - name: "tenant_id"
+    type: "string"
+    required: true
+    credential: false
+    description: |
+      Your Office 365 tenant ID. This is ID is located on the **Properites** page in your Azure Active Directory.
+    value: "<YOUR_TENANT_ID>"
```
erinkcochran87, line 64: You're missing the closing dashes (`---`) on this file. If you look at [this preview of the Source & Destination API Availability reference](https://deploy-preview-575--stitchdocs.netlify.app/docs/developers/stitch-connect/guides/source-destination-reference#sources-api-availability), you'll see that Microsoft Teams is listed in the section for the sources, but that the **API connection property** column is blank. That's because Jekyll didn't render this file, because the dashes are missing.

erinkcochran87 (changes requested):
Back to you!

_integration-schemas/ms-teams/v1/foreign-keys.md
```diff
@@ -0,0 +1,153 @@
...
+#     USING THIS TEMPLATE    #
+# -------------------------- #
+
+## NEED HELP USING THIS TEMPLATE? SEE:
+## https://docs-about-stitch-docs.netlify.com/reference/integration-templates/saas-foreign-keys/
+## FOR INSTRUCTIONS & REFERENCE INFO
+
+tap-reference: "integration"
+
+version: ""
```
erinkcochran87, line 12: I'm gonna fix this because I want to do a few things in here, but remember to double-check that these two guys are correct or all the FK work you do won't show up

erinkcochran87 (approved):

PR #583, opened 2020-11-11
Chargify/v1

erinkcochran87:
This PR adds documentation for the new Chargify integration.


erinkcochran87:
@jazsmi There isn't an icon for this integration yet, but everything else is done.

PR #591, opened 2020-12-02
MS SQL - Server Schema sections

jazsmi:
Adds `Include Microsoft SQL Server schema names in destination tables (uncommon)` to the MS SQL docs.


erinkcochran87 (changes requested):
I think there's an opportunity to single source this content a bit more - see JIRA for my comments

erinkcochran87 (approved):

PR #608, opened 2021-02-11
Pendo - update rep keys

jazsmi:
change `guides` and `poll_events` tables' replication keys to `browserTime` (source: https://github.com/singer-io/tap-pendo/pull/21)

### DIFF
--- a/_integration-schemas/pendo/v1/guides.md
+++ b/_integration-schemas/pendo/v1/guides.md
@@ -10,8 +10,6 @@ description: |
   The `{{ table.name }}` table contains info about the guides in your {{ integration.display_name }} account.
 
 replication-method: "Key-based Incremental"
-replication-key:
-  name: "browserTime"
 
 api-method:
   name: "Aggregation"
@@ -27,6 +25,7 @@ attributes:
   - name: "last_updated_at"
     type: "date-time"
     description: ""
+    replication-key: true
 
   - name: "attributes"
     type: "object"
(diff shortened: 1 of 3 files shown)


_integration-schemas/pendo/v1/poll_events.md
```diff
@@ -13,7 +13,7 @@ description: |
 
 replication-method: "Key-based Incremental"
 replication-key:
```
erinkcochran87, line 15: This is actually an attribute that will be in the table. `replication-key.name` as you have it here should only be used when this is *not* the case. 

[Here's the schema ](https://github.com/bytecodeio/tap-pendo-release/blob/05a2bd778230fba39d1a9671004ef9d7efb2418f/tap_pendo/schemas/poll_events.json#L27)for `poll_events`.

Also note that it's actually `browser_time`.

_integration-schemas/pendo/v1/guides.md
```diff
@@ -10,6 +10,8 @@ description: |
   The `{{ table.name }}` table contains info about the guides in your {{ integration.display_name }} account.
 
 replication-method: "Key-based Incremental"
+replication-key:
```
erinkcochran87, line 13: It looks like the ticket I wrote that corresponds to this was written badly, but the Rep Key change is actually for the `guide_events` table, not `guides`. I recommend always checking the pull request that corresponds with our changes to make sure you're working on the right thing.

[See the PR here](https://github.com/singer-io/tap-pendo/pull/21/files#diff-c228b69d34f7069450ae23364f654671e3c0ba6502c989b4aca087d37ae622bb)

Also, `browserTime` is `browser_time` and it's an attribute in the actual table.

erinkcochran87 (changes requested):

erinkcochran87 (approved):

PR #619, opened 2021-03-17
eBay - 90 historical window

jazsmi:
Mentioned in the setup and and `orders` table to set a start date within 90 days.

### DIFF
--- a/_integration-schemas/ebay/v1/orders.md
+++ b/_integration-schemas/ebay/v1/orders.md
@@ -7,7 +7,7 @@ name: "orders"
 doc-link: "https://developer.ebay.com/api-docs/sell/fulfillment/resources/order/methods/getOrders"
 singer-schema: "https://github.com/singer-io/tap-ebay/blob/master/tap_ebay/schemas/orders.json"
 description: |
-  The {{ table.name }} table contains details about all orders in your {{ integration.display_name }} seller account. This table must use a request with a Start Date set that is within 90 days.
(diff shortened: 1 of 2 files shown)


_integration-schemas/ebay/v1/orders.md
```diff
@@ -7,7 +7,7 @@ name: "orders"
 doc-link: "https://developer.ebay.com/api-docs/sell/fulfillment/resources/order/methods/getOrders"
 singer-schema: "https://github.com/singer-io/tap-ebay/blob/master/tap_ebay/schemas/orders.json"
 description: |
-  The {{ table.name }} table contains details about all orders in your {{ integration.display_name }} seller account.
+  The {{ table.name }} table contains details about all orders in your {{ integration.display_name }} seller account. This table must use a request with a Start Date set that is within 90 days.
```
erinkcochran87, line 10: What error would the user receive if they set it beyond 90 days? Include that here so they can identify the problem when they receive the error.

_saas-integrations/ebay/v1/ebay-v1.md
```diff
@@ -99,6 +99,8 @@ setup-steps:
     content: |
       {% include integrations/saas/setup/historical-sync.html %}
 
+      To retreive data from the `orders` table, the historical start date must be within 90 days.
```
erinkcochran87, line 102:
```suggestion
      **Note**: To retrieve data from the `orders` table, the **Start Date** must not exceed 90 days from the current date, whether you're initially creating the integration or resetting this parameter. If this exceeds 90 days, you'll encounter [errors during Extraction](#orders).
```

This clarifies what exactly we mean when we say `90 days`. It also links right to the documentation for the affected table, where you'll include the error message the user would receive if they didn't heed the warning. :)

erinkcochran87 (changes requested):

erinkcochran87 (approved):

PR #629, opened 2021-04-13
MySQL/v2

jazsmi:
High Performance version


_developer-files/connect/api/objects/form-properties/sources/databases/hp-mysql-object.md
```diff
@@ -0,0 +1,139 @@
...
+
+# -------------------------- #
+#      OBJECT ATTRIBUTES     #
+# -------------------------- #
+
+uses-common-fields: true
+uses-feature-fields: true
+uses-start-date: true
+
+object-attributes:
```
erinkcochran87, line 50: - `allow_non_auto_increment_pks` - It doesn't look like this is in the report card for `tap.hp-mysql`. I'm not sure if this is intentional or not, but it should be confirmed with the sources team and then removed if needed

- `ssh`-related properties, `server_id`, `ssl_client_auth_enabled` are all displaying twice. I think this is because you have them listed out here, but they don't need to be. The `uses-feature-fields` flag being set to `true` should make it so that you don't need to do that.

erinkcochran87 (changes requested):
Other things I noticed, but can't comment on since there aren't files to do so yet:

- The version file needs to be updated to include v2
- Changelog entry

erinkcochran87 (approved):

PR #644, opened 2021-05-18
Shopify - Shopify Plus permissions

jazsmi:
This PR adds info about possible permission requirements for Shopify Plus plan users to the Shopify docs.

### DIFF
--- a/_integration-schemas/shopify/customers.md
+++ b/_integration-schemas/shopify/customers.md
@@ -6,7 +6,7 @@ name: "customers"
 doc-link: "https://shopify.dev/docs/admin-api/rest/reference/customers"
 singer-schema: "https://github.com/singer-io/tap-shopify/blob/master/tap_shopify/schemas/customers.json"
 description: |
-  The `{{ table.name }}` table contains info about the shop's customers. This includes their contact details, order history, and email marketing preferences. Null values are now accepted for replication.
+  The `{{ table.name }}` table contains info about the shop's customers. This includes their contact details, order history, and email marketing preferences.
 
   #### Customer metafield data
 
--- a/_integration-schemas/shopify/customers.md
+++ b/_integration-schemas/shopify/customers.md
@@ -6,7 +6,7 @@ name: "customers"
 doc-link: "https://shopify.dev/docs/admin-api/rest/reference/customers"
 singer-schema: "https://github.com/singer-io/tap-shopify/blob/master/tap_shopify/schemas/customers.json"
 description: |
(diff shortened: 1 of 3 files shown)


_saas-integrations/shopify/v1/shopify-v1.md
```diff
@@ -66,6 +66,9 @@ feature-summary: |
 #      Setup Instructions    #
 # -------------------------- #
 
+requirements-info: |
```
erinkcochran87, line 69: This isn't displaying on the preview page - did you test it?

Hint: Take a look at the `singer.html` layout to see why.

_saas-integrations/shopify/v1/shopify-v1.md
```diff
@@ -66,6 +66,9 @@ feature-summary: |
...
+  Staff members on a {{ integration.display_name }} Plus plan cannot export data by default. Store owners can give staff members permissions export orders, draft orders, products, inventory, and customer data. Read the [{{ integration.display_name }} Staff permissions documentation](https://help.shopify.com/en/manual/your-account/staff-accounts/staff-permissions#store-owner-permissions){:target="new"} for more information.
```
erinkcochran87, line 70: Consider re-arranging some of this so that it matches the structure of requirements in other integrations. By doing this, you can also make it clear in the first sentence that the requirement is applicable only to a subset of Shopify users:

```
**If you're on a {{ integration.display_name }} Plus plan**, [...remaining content...]
```

_integration-schemas/shopify/customers.md
```diff
@@ -6,7 +6,7 @@ name: "customers"
 doc-link: "https://shopify.dev/docs/admin-api/rest/reference/customers"
 singer-schema: "https://github.com/singer-io/tap-shopify/blob/master/tap_shopify/schemas/customers.json"
 description: |
-  The `{{ table.name }}` table contains info about the shop's customers. This includes their contact details, order history, and email marketing preferences.
+  The `{{ table.name }}` table contains info about the shop's customers. This includes their contact details, order history, and email marketing preferences. Null values are now accepted for replication.
```
erinkcochran87, line 9: I think we updated the changelog with this last month: https://github.com/singer-io/tap-shopify/pull/94/files

I don't think we need to explicitly mention this. Typically we'll allow `null` values for replication, with the exceptions potentially being Primary and Replication Keys.

erinkcochran87 (changes requested):

erinkcochran87:
@jazsmi Could you also please add a changelog entry for the permissions update? That would be a `documentation` update - definitely worth calling out since it may affect a fair number of users.

erinkcochran87 (approved):

PR #746, opened 2022-03-23
Fix some activecampaign schemas

luandy64:
This PR is the result of me running a branch of `tap-generate-docs` for `tap-activecampaign`.

A lot of the changes are a `null` to `number` update. This seems to be because the tap has schemas that look like

```
"price": {
      "anyOf": [
          {
              "type": "null"
          },
          {
              "type": "number",
              "multipleOf": 1e-8
          }
      ]
    },
```
[Source](https://github.com/singer-io/tap-activecampaign/blob/a5b5656b989ba684b275a1eb36bbc95afa2f2855/tap_activecampaign/schemas/ecommerce_order_products.json#L23-L33)

when it really should just be

```
"price": {
     "type": ["null", "number"],
     "multipleOf": 1e-8
},
```

---

#### Change 1
[Here is the tap using a schema of an empty array and object](https://github.com/singer-io/tap-activecampaign/blob/a5b5656b989ba684b275a1eb36bbc95afa2f2855/tap_activecampaign/schemas/automation_blocks.json#L14-L30)

This PR just makes it a
```
type: "array, object"
description: ""
subattributes: [
]
```

#### Change 2

A handful of fields have the name slightly changed. I'm not sure which is correct, but I included them in this PR so that we can review them. There's always a small chance I missed a field or two as well, especially since there are 60 streams in this tap.


lmilan:
The display of empty arrays/objects is a bit weird, there is a "Click to expand" that doesn't do anything (see `params` in `[automation_blocks](https://deploy-preview-746--stitchdocs.netlify.app/docs/integrations/saas/activecampaign#automation-blocks)` for example), should we just remove the subattributes element in this case?

luandy64:
I agree that's confusing. I can update `tap-generate-docs` to remove subattributes in that case

lmilan (approved):

PR #889, opened 2023-03-21
TDL-22391 tap invoiced add plans stream

dmosorast:
# Description of change
Adding the plans stream to Invoiced. This stream was added by a community contributor in version 0.2.0.

# Manual QA steps
 - Generated with the tool and reviewed with Andy to fill in the rest of the text
 
# Risks
 - Some of these fields don't appear in the Invoiced docs, but the PR was reviewed by the original developer (an Invoiced employee), so I'm trusting that.
 
# Rollback steps
 - revert this branch

### DIFF
--- a/_integration-schemas/invoiced/plans.md
+++ b/_integration-schemas/invoiced/plans.md
@@ -33,6 +33,7 @@ attributes:
   - name: "id"
     type: "string"
     description: "The plan’s unique ID"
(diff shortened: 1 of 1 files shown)


lmilan (commented):
Hi, we don't have the replication key or primary key specified, should it be `updated_at` and `id` like the other streams?

